PR #19, opened 2022-10-12
feat: Initial Gomobile interface definitions

### LINKED ISSUE #17: VC Wallet Library - Go mobile/web interfaces
Add interfaces that are similar to the ones in #15, but are modified to be compatible with `gomobile` bindings.

DRK3:
Added interface definitions that are similar to the existing ones, but modified to be compatible with gomobile. These interface definitions will likely be iterated on and changed in the future as development work continues.

closes #17


DRK3:
In my next PRs, I will:
* Add script(s) to generate the bindings
* Add stub implementations for @talwinder50 to use

sudeshrshetty (approved):

PR #21, opened 2022-10-14
feat: Gomobile SDK implementation stubs

DRK3:
Added stubs for various SDK functionality. This is intended as a simple starting point to allow for testing of the gomobile bindings. Real functionality still needs to be added, and the method signatures and comments/documentation still need to be worked on. Many of them are just placeholders.

Some other notes:
* Added a Crypto interface.
* For the storage implementation stubs, I added very basic storage functionality that we can try using while testing out the gomobile bindings.
* Updated linter script to lint the wallet-sdk-gomobile folder.

### DIFF
--- a/cmd/wallet-sdk-gomobile/api/api.go
+++ b/cmd/wallet-sdk-gomobile/api/api.go
@@ -7,12 +7,14 @@ SPDX-License-Identifier: Apache-2.0
 // Package api defines gomobile-compatible wallet-sdk interfaces.
 package api
 
+// KeyHandle represents a key with associated metadata.
 type KeyHandle struct {
 	Key     []byte `json:"key,omitempty"`
 	KeyType string `json:"keyType,omitempty"`
-	KeyID   string `json:"keyID,omitempty"`
+	KeyID   string `json:"keyID,omitempty"` //nolint: tagliatelle //false positive
 }
 
+// A KeyHandleWriter represents a type that is capable of performing certain operations related to writing key data.
 type KeyHandleWriter interface {
 	// Create a new key/keyset/key handle of type keyType
 	// Some key types may require additional attributes described in `opts`. // TODO: Format of opts to be determined.
@@ -25,7 +27,8 @@ type KeyHandleWriter interface {
 	Rotate(keyType, keyID string, opts string) (*KeyHandle, error)
 	// Import will import privKey into the KMS storage for the given keyType then returns the new key id and
 	// the newly persisted KeyHandle.
-	// privKey possible types are: *ecdsa.PrivateKey and ed25519.PrivateKey. // TODO: Determine how these restrictions work.
+	// privKey possible types are: *ecdsa.PrivateKey and ed25519.PrivateKey.
+	// TODO: Determine how these restrictions work.
 	// kt possible types are signing key types only (ECDSA keys or Ed25519)
 	// opts allows setting the keysetID of the imported key using WithKeyID() option. If the ID is already used,
 	// then an error is returned. // TODO: Format of opts to be determined.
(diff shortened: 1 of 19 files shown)


DRK3:
Note that the Java bindings don't work if there's a type with the same name as the package it's in. This necessitated me changing around some of the names - some are good, but some are probably a bit strange. If you have better name suggestions let me know - otherwise, I'll figure out better names as I work on the implementations.

pkg/api.go
```diff
@@ -85,6 +87,97 @@ type CredentialWriter interface {
...
+	// 		cipherText in []byte
+	//		nonce in []byte
+	//		error in case of errors during encryption
+	Encrypt(msg, aad []byte, kh interface{}) ([]byte, []byte, error)
+	// Decrypt will decrypt cipher with aad and given nonce using a matching AEAD primitive in kh key handle of a
+	// private key
+	// returns:
+	//		plainText in []byte
+	//		error in case of errors
+	Decrypt(cipher, aad, nonce []byte, kh interface{}) ([]byte, error)
```
rolsonquadras, line 103: @sudeshrshetty @DRK3 Do we need Encrypt/Decrypt/DeriveProof/Blind for OpenID flows ? We can start with minimum and then add functions.

DRK3: Just to confirm: do we need all of the methods _except_ for those specific 4 ones you listed? Or is there a smaller subset?

rolsonquadras: For now, just keep Sign/Verify.

DRK3: Done

rolsonquadras (approved):

cmd/wallet-sdk-gomobile/api/api.go
```diff
@@ -25,56 +27,79 @@ type KeyHandleWriter interface {
...
-	//  - error if it fails to export the public key bytes
+	//  - Error if it fails to export the public key bytes
 	Export(keyID string) (*KeyHandle, error)
 }
 
+// DIDCreator defines the method required for a type to create DID documents.
 type DIDCreator interface {
 	// Create creates a new DID Document.
 	// It returns a DID Document Resolution.
 	Create(didDocument []byte) ([]byte, error)
```
talwinder50, line 61: So who is passing did document here ?

sudeshrshetty, line 61: didDocument as request??

talwinder50 (approved):

cmd/wallet-sdk-gomobile/api/api.go
sudeshrshetty: https://github.com/trustbloc/agent-sdk/blob/main/cmd/wallet-js-sdk/src/did/didmanager.js#L71

DRK3: Updated the PR. I change the method signature to take in a `CreateDIDOpts` in a similar style as some of the other various methods in the SDK introduced by this PR. I only added a `Method` option for now since I'm not sure exactly what will be required for the various DID methods - when I add a `did:Key` DIDCreator implementation I can update this as needed. (Or if you have a good idea of what exactly we'll need, let me know and I'll add it now)

DRK3: Take a look at the thread here and let me know if you still have a question: https://github.com/trustbloc/wallet-sdk/pull/21#discussion_r998591087

cmd/wallet-sdk-gomobile/credentialinquirer/credentialinquirer.go
```diff
@@ -0,0 +1,48 @@
...
+	// used in the constructor (NewInquirer).
+	VCs []byte
+	// CredentialReader allows for access to a VC storage mechanism. If specified, this takes precedence over the
+	// CredentialReader used in the constructor (NewInquirer).
+	CredentialReader api.CredentialReader
+}
+
+// Query returns credentials that match the given query.
+// Credentials are returned as a JSON array.
+func (*Inquirer) Query(queryOpts QueryOpts) ([]byte, error) {
```
sudeshrshetty, line 46: it should have 2 params, 

- `query` - string
-  `data` - VCs

cmd/wallet-sdk-gomobile/credentialschema/credentialschema.go
```diff
@@ -0,0 +1,47 @@
...
+	// used in the constructor (NewResolver).
+	VCs []byte
+	// CredentialReader allows for access to a VC storage mechanism. If specified, this takes precedence over the
+	// CredentialReader used in the constructor (NewResolver).
+	CredentialReader api.CredentialReader
+}
+
+// Resolve resolves the given credentials and returns resolved descriptors. See ResolveOpts for more information.
+// Returns a JSON array of resolved descriptors.
+func (c *Resolver) Resolve(resolveOpts ResolveOpts) ([]byte, error) {
```
sudeshrshetty, line 45: you have to split arguments - 

- schema [manifest OR credentialSuppported]
- Data to resolve - single VC or Credential ID(credential reader has to be present to work with credential ID)

sudeshrshetty (approved):

cmd/wallet-sdk-gomobile/credentialinquirer/credentialinquirer.go
DRK3: Done

cmd/wallet-sdk-gomobile/credentialschema/credentialschema.go
DRK3: Done

PR #30, opened 2022-10-21
feat: Create did:key method

DRK3:
* Added a did:key creator implementation.
* Updated README with additional information and clarification on the project structure.
* Updated linter.
* Updated Android SDK version used for Android gomobile bindings to something a little newer but still supported by most devices (22 instead of 19).
* Fixed issues where some methods used types that weren't compatible with gomobile, causing those methods to be skipped during binding generation.


DRK3:
Important info about this PR, please read:

The "pure-Go" SDK is in the `pkg` folder. This is meant to be used by a user who wants to write a native Go application.

In the `cmd/wallet-sdk-gomobile` folder I have similar but slightly different version of the SDK. This one acts as a wrapper for the Go-SDK. It converts between `gomobile`-compatible types internally and also calls the "pure-Go" SDK internally. See the updated README for more info.

Because the Go SDK and `gomobile`-compatible SDK each are separate Go modules, the Go SDK did:key creator and `gomobile`-compatible versions have to be in separate PRs. For the sake of showing what this project structure looks like, I have them both in one PR for this initial draft, but before finalizing it I will be removing the `gomobile` version from this PR and creating a followup PR with it.

cmd/wallet-sdk-gomobile/api/api.go
```diff
@@ -58,13 +58,14 @@ type KeyHandleReader interface {
 type CreateDIDOpts struct {
 	// The type of DID to create (i.e. which DID method to use).
 	Method string
+	Key    []byte
 }
 
 // DIDCreator defines the method required for a type to create DID documents.
 type DIDCreator interface {
```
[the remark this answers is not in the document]
DRK3: In other words - can we really define one type of "create" method that works for all DID methods...?

cmd/wallet-sdk-gomobile/did/creator/creator.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// NewDIDCreator returns a new DIDCreator.
+func NewDIDCreator() *DIDCreator {
+	return &DIDCreator{goAPIDIDCreator: goapididcreator.NewDIDCreator()}
+}
+
+// Create creates a DID document for the given DID method.
+// The usage of createDIDOpts depends on the DID method you're using.
+func (d *DIDCreator) Create(method string, createDIDOpts *api.CreateDIDOpts) ([]byte, error) {
```
DRK3, line 33: We should discuss how the caller is expected to pass their key in. In this PR, I have them pass in the public key directly, assuming they generated it somehow (maybe with the KMS implementation, to be done in a future PR?)

sudeshrshetty, line 33: There is a `method` already inside `CreateDIDOpts`

cmd/wallet-sdk-gomobile/api/api.go
```diff
@@ -58,13 +58,14 @@ type KeyHandleReader interface {
 type CreateDIDOpts struct {
```
sudeshrshetty, line 58: please follow this pattern : https://github.com/trustbloc/agent-sdk/blob/main/cmd/wallet-js-sdk/src/did/didmanager.js#L71-L181

pkg/api/api.go
```diff
@@ -58,10 +59,16 @@ type KeyHandleReader interface {
 	Export(keyID string) ([]byte, kms.KeyType, error)
 }
 
-// DIDCreator defines DID creation APIs.
+// CreateDIDOpts represents the various options for the DIDCreator.Create method.
+type CreateDIDOpts struct {
+	DIDDoc *did.Doc
```
sudeshrshetty, line 64: What is this DIDDoc option?

cmd/wallet-sdk-gomobile/did/creator/creator.go
DRK3: Oops, I meant to remove that. Fixed

pkg/api/api.go
DRK3: The various Aries DID creator methods (which I call from wallet-sdk here) require this as an input parameter (see https://github.com/hyperledger/aries-framework-go/blob/a755fd87d699b886341536c690dd78616cf623bd/pkg/vdr/key/creator.go#L31 for reference).

Is seems the Aries DID creator methods use the fields from that DID Doc input parameter as options, essentially. Not sure why they work like that. Maybe @Baha-sk @fqutishat have some idea?

Perhaps I'll have to define those "options" as input parameters for our wallet-sdk function, and then inside our wallet-sdk function I can translate them to the "DID Doc" input parameter?

cmd/wallet-sdk-gomobile/api/api.go
DRK3: I need some clarification on which of those parameters applies here, since that `didmanager.js` method is specific to `did:orb`, while this `CreateDIDOpts` is more generic. Also, please see here: https://github.com/trustbloc/wallet-sdk/pull/30#discussion_r1008310488 I also need some clarification on how this DIDCreator interface should work.

pkg/did/creator/creator.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/trustbloc/wallet-sdk/pkg/api"
+	"github.com/trustbloc/wallet-sdk/pkg/did/key"
+)
+
+// DIDCreator is used for creating DID Documents using supported DID methods (currently only did:key).
+type DIDCreator struct{}
+
+// NewDIDCreator returns a new DIDCreator.
+func NewDIDCreator() *DIDCreator {
```
DRK3, line 23: Pass in KeyHandleWriter (create key pair if none specified) and KeyHandleReader (both optional)

cmd/wallet-sdk-gomobile/api/api.go
```diff
@@ -56,15 +56,14 @@ type KeyHandleReader interface {
 
 // CreateDIDOpts represents the various options for the DIDCreator.Create method.
 type CreateDIDOpts struct {
-	// The type of DID to create (i.e. which DID method to use).
-	Method string
+	Key []byte
```
talwinder50, line 59: I think key name is bit confusing may be keyType ??

cmd/wallet-sdk-gomobile/api/api.go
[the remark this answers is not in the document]
DRK3: After some discussion, we will re-evaluate later when we add another DID method

cmd/wallet-sdk-gomobile/did/creator/creator.go
DRK3: Updated based on a discussion earlier

cmd/wallet-sdk-gomobile/api/api.go
DRK3: Updated based on our discussion earlier.

cmd/wallet-sdk-gomobile/api/api.go
DRK3: The `Key` there is the raw key bytes. However, I've updated my PR so that's no longer there - take a look and let me know what you think!

cmd/wallet-sdk-gomobile/did/creator/creator.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (g *gomobileKHRWrapper) Export(keyID string) ([]byte, error) {
+	return nil, errors.New("not implemented")
+}
+
+// NewDIDCreator returns a new DIDCreator.
+// keyHandleReader is optional. If provided, then it will be used to fetch a key (based on the KeyID in the options)
+// in the Create method. If not provided, then a new key will be generated and used for DID creation.
+func NewDIDCreator(keyHandleReader api.KeyHandleReader) *DIDCreator {
```
DRK3, line 52: We talked earlier about having both a key handle reader and writer as optional parameters, and the key handle writer would be used to write an automatically generated key into. However, while working on this I realized an issue here - what KeyID would I use? How would the caller know what KeyID I used? For now, I left it out since I couldn't quite make sense of what to do there, but we can discuss this more

cmd/wallet-sdk-gomobile/did/creator/creator.go
```diff
@@ -0,0 +1,73 @@
...
+	goAPIKHRWrapper := &gomobileKHRWrapper{keyHandleReader: keyHandleReader}
+
+	return &DIDCreator{goAPIDIDCreator: goapididcreator.NewDIDCreator(goAPIKHRWrapper)}
+}
+
+// Create creates a DID document for the given DID method.
+// The usage of createDIDOpts depends on the DID method you're using.
+func (d *DIDCreator) Create(method string, createDIDOpts *api.CreateDIDOpts) ([]byte, error) {
```
DRK3, line 60: For now, the only options I expose through my API here are: key source (auto-generated or passed in via keyHandleReader) and the key type (if using keyHandleReader, this is determined by the stored key type). However, the did_key Creator functionality in Aries does have more optional settings. Should I expose more of them? Not sure how complicated we want this API here to be.

cmd/wallet-sdk-gomobile/did/creator/creator.go
```diff
@@ -0,0 +1,73 @@
...
+
+	didDocResolution, err := d.goAPIDIDCreator.Create(method, goAPIOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	return didDocResolution.JSONBytes()
+}
+
+func convertToGoAPIOpts(createDIDOpts *api.CreateDIDOpts) *goapi.CreateDIDOpts {
```
DRK3, line 71: Conversions like this may seem funny since the two struct types have identical members, but the reason I have to do this is that `gomobile` doesn't allow me to use types defined from packages outside of the packages I'm generating bindings on (it skips generating those methods)

cmd/wallet-sdk-gomobile/api/api.go
```diff
@@ -11,7 +11,6 @@ package api
 type KeyHandle struct {
 	Key     []byte `json:"key,omitempty"`
```
sudeshrshetty, line 12: you may have to define encoding if it is keybytes

cmd/wallet-sdk-gomobile/did/creator/creator.go
sudeshrshetty: Caller will know looking at your DIDDocument about KID of key pairs created

pkg/did/creator/creator.go
```diff
@@ -0,0 +1,77 @@
...
+func NewDIDCreator(keyHandleReader api.KeyHandleReader) *DIDCreator {
+	return &DIDCreator{
+		keyHandleReader: keyHandleReader,
+	}
+}
+
+// Create creates a DID document for the given DID method.
+// The usage of createDIDOpts depends on the DID method you're using.
+func (d *DIDCreator) Create(method string, createDIDOpts *api.CreateDIDOpts) (*did.DocResolution, error) {
+	if method == DIDMethodKey || method == fmt.Sprintf("%s%s", didMethodPrefix, DIDMethodKey) {
```
sudeshrshetty, line 45: I didn't quite the get the second part of OR condition

pkg/did/creator/creator.go
```diff
@@ -0,0 +1,77 @@
...
+
+	return nil, fmt.Errorf("DID method %s not supported", method)
+}
+
+func (d *DIDCreator) createKeyDIDDoc(createDIDOpts *api.CreateDIDOpts) (*did.DocResolution, error) {
+	var key []byte
+
+	var keyType string
+
+	if d.keyHandleReader == nil {
```
sudeshrshetty, line 57: Can you explain this?

When client provided `kid` then you simply read from key reader & if key reader not setup then throw error.

By default, you create a keypair in keyhandle writer using key type provided.
Use defauult ED25519 as keytype.

cmd/wallet-sdk-gomobile/api/api.go
DRK3: Added a comment to clarify that it's raw key bytes.

pkg/did/creator/creator.go
DRK3: This just allows someone to use either "key" or "did:key" as the method name.

pkg/did/creator/creator.go
DRK3: What the code here does is, if the key reader is set up, then I try reading from there. Otherwise, I generate a new key for them and use that for the did document creation. I'll update it so that if the keyID is used but no key reader is set up, then an error gets thrown. I haven't added a keyHandleCreator yet due to the reason in my other comment (will discuss with you separately, as that one is a bit more complicated)

rolsonquadras (approved):

pkg/did/creator/creator.go
sudeshrshetty: mobile user's don't care, it is not an actual VDR. To simplify keep it simple as `key`

pkg/api/api.go
sudeshrshetty: delete it, we dont need it here. In aries we follow some opensource specs & some protocols where we had to use some custom data types or interfaces.

Try to keep wallet SDK as simple as possible

sudeshrshetty (approved):

pkg/did/creator/creator.go
DRK3: Done in https://github.com/trustbloc/wallet-sdk/pull/46

PR #32, opened 2022-10-25
feat: Flutter project setup and embedded walletsdk framework

talwinder50:
Screen Shot 2022-10-25 at 6 26 21 PM
- Flutter Project Setup 
- Hello Function to make sure flutter app works 
- Created Plugin to embed the wallet sdk framework 
- Used StoreNewProvider to store dummy credential (You can open the project in xcode workspace and run ) 
- Add the make target to run the app on ios simulator iphone 14 
- Deleted current capacitor app under demo not required 

This is the initial push will lead to refine this in the following pr . 


README.md
```diff
@@ -16,12 +16,12 @@ This project can be used as a
...
   - Android (generates a .aar binary)
 
-The repo also has code to generate [Reference iOS/Android App](./demo/app/) built using [Ionic](https://ionicframework.com/)/[Capacitor JavaScript](https://capacitorjs.com/) frameworks.
+The repo also has code to generate [Reference iOS/Android App](demo/app_1/) built using [Ionic](https://ionicframework.com/)/[Capacitor JavaScript](https://capacitorjs.com/) frameworks.
 
 
 ## Build/Run
 - [GoMobile Bindings (iOS/Android)](./cmd/wallet-sdk-gomobile/README.md)
-- [Demo/Reference App](./demo/app/README.md)
+- [Demo/Reference App](demo/app_1/README.md)
```
rolsonquadras, line 24: Should this be reverted back to `/demo/app`

demo/app/README.md
```diff
@@ -1,10 +1,20 @@
-# Demo/Reference App
+# wallet_sdk_flutter
```
rolsonquadras, line 1: Can we revert the title to existing one ?

demo/app/README.md
```diff
@@ -1,10 +1,20 @@
...
 
-TODO: Add introduction
+A new Flutter project.
```
rolsonquadras, line 3: Looks like we need to add details/intro here. Can we keep the TODO ?

rolsonquadras (approved):
Few minor comments

demo/app/README.md
talwinder50: fixed

demo/app/ios/Runner/flutterPlugin.swift
```diff
@@ -0,0 +1,43 @@
...
+          
+          let storageProvider = StorageNewProvider()
+          do {
+              let jsonVC = """
+              {
+                "id": "http://example.edu/credentials/1872"
+              }
+              """.data(using: .utf8)!
+              try storageProvider!.add(jsonVC)
+              print("credential is stored")
```
vkubiv, line 30: you should call here result() too. in another case, the promise in flutter part will not complete. You can call it with out arguments.

demo/app/ios/Runner/flutterPlugin.swift
```diff
@@ -0,0 +1,43 @@
...
+          } catch {
+              result("error while storing verfiable credential")
```
vkubiv, line 32: wrap error message in special error class FlutterError.

Like this:
result(FlutterError.init(code: "NATIVE_ERR",
                                             message: "error while storing verfiable credential",
                                             details: nil))

You can check here https://stackoverflow.com/questions/50078947/how-to-implement-a-flutterplugins-method-handler-in-swift

demo/app/ios/Runner/flutterPlugin.swift
```diff
@@ -0,0 +1,43 @@
...
+
+public class SwiftHelloPlugin: NSObject, FlutterPlugin {
+  public static func register(with registrar: FlutterPluginRegistrar) {
+    let channel = FlutterMethodChannel(name: "HelloPlugin", binaryMessenger: registrar.messenger())
+    let instance = SwiftHelloPlugin()
+    registrar.addMethodCallDelegate(instance, channel: channel)
+  }
+ 
+  public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+      if (call.method == "storeCredentials"){
```
vkubiv, line 20: One suggestion for future: do not directly put code inside if's in handle function. Better to like this:

```
public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
      if (call.method == "storeCredentials"){
         storeCredentials(args["name"] as! String, result)
      }
}

public func storeCredentials(_ someArg: String, result: @escaping FlutterResult) {
 //Code will be here
}
```

vkubiv (approved):

PR #43, opened 2022-11-02
feat: Andriod Plugin and Integrate did key

talwinder50:
Screen Shot 2022-11-02 at 12 13 46 PM
Walletsdk.xcframework and wallet.aar (ios and andriod bindings ) wont be pushed in the future pr after target of automatically setting them up is setup in place 

https://user-images.githubusercontent.com/7862595/199543172-b4fe20c7-df97-4d67-9d5f-8618e3d39c7c.mp4

closes #42

Signed-off-by: Talwinder kaur <talwinder.kaur@avast.com>

PR #44, opened 2022-11-02
chore: Add android/iOS binaries to gitignore

rolsonquadras:
Signed-off-by: Rolson Quadras <rolson.quadras@avast.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,5 @@ deployments/demo/openapi/specs
 
 # Built gomobile bindings
 cmd/wallet-sdk-gomobile/bindings
+*.aar
+*.xcframework


.gitignore
```diff
@@ -38,3 +38,5 @@ deployments/demo/openapi/specs
 
 # Built gomobile bindings
 cmd/wallet-sdk-gomobile/bindings
+*.aar
```
DRK3, line 41: These should already be excluded with the line above.

rolsonquadras: The line above ignores whatever inside `/bindings/` folder. These new lines ignores binaries in all places inside the repo/project.

DRK3: Why would the binaries be inside the repo otherwise?

DRK3: Do they get downloaded there by the package manager for the reference implementation?

rolsonquadras: To avoid instances like https://github.com/trustbloc/wallet-sdk/pull/43#issuecomment-1300820908 and https://github.com/trustbloc/wallet-sdk/pull/43#issuecomment-1300823243

DRK3 (approved):

talwinder50 (approved):

PR #72, opened 2022-11-17
feat: In-memory credential storage implementation

DRK3:


DRK3:
Note: I tried implementing the `JSONObject` and `JSONArray` types using two other alternative means:

Type alias: `type JSONObject = string`. This didn't work because `gomobile` automatically converts them to strings, so the name is lost (defeating the purpose).

String type definition: `type JSONObject string`. `gomobile` just doesn't support doing this.

vkubiv (approved):

cmd/wallet-sdk-gomobile/api/api.go
```diff
@@ -7,6 +7,20 @@ SPDX-License-Identifier: Apache-2.0
 // Package api defines gomobile-compatible wallet-sdk interfaces.
 package api
 
+// JSONObject contains a single JSON object (not an array).
+// It's a simple wrapper around the actual JSON string. Its purpose is to help the
+// caller using the mobile bindings to understand what type of data to expect or pass in.
+type JSONObject struct {
+	JSON string
```
sudeshrshetty, line 14: so we decided to have string here instead of bytes?

bytes are ideal way to pass large data.

sudeshrshetty: rename `JSON` to `data`, struct name already implies that it is JSON

DRK3: Renamed

DRK3: This can help differentiate it from something key bytes, which truly are raw bytes

DRK3: One reason I named it `JSON` instead of `Data` is because I didn't want the caller to confuse the struct itself with the actual JSON string... although I'm not sure if there's a super clean way to do that.

DRK3: Changed back to []byte for now

sudeshrshetty (approved):

PR #76, opened 2022-11-23
feat: Add flutter secure storage feature for credential list and credential preview 

### LINKED ISSUE #77: Implement Flutter secure storage

talwinder50:
https://user-images.githubusercontent.com/7862595/203448352-70261a49-5402-4a44-9826-af68d2f4236e.mp4

IMG_0153
Screenshot 2022-11-22 at 6 47 07 PM

closes #77


rolsonquadras (approved):

DRK3 (approved):
One small note: what you're storing is the full credential response from the OpenID4CI flow, which has the credential in base64 format along with some other data. Not sure if we need to store the whole response or not, and we should consider decoding the base64 data for storage efficiency reasons, but I guess this isn't a huge priority at the moment (perhaps we can look at it later)

PR #79, opened 2022-11-24
feat: aries crypto to work with local kms.

vkubiv:


cmd/wallet-sdk-gomobile/credentialquery/credentialquery.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/piprate/json-gold/ld"
 
 	"github.com/trustbloc/wallet-sdk/cmd/wallet-sdk-gomobile/api"
-	"github.com/trustbloc/wallet-sdk/internal/gomobilewrappers"
+	"github.com/trustbloc/wallet-sdk/cmd/wallet-sdk-gomobile/gomobilewrappers"
```
DRK3, line 20: No longer internal? Was there some issue you ran into? Those wrappers will appear in the bindings if it's not internal

cmd/wallet-sdk-gomobile/localkms/localkms.go
```diff
@@ -18,15 +18,27 @@ import (
 // KeyTypeED25519 is the name recognized by the Create method for creating an ED25519 keyset.
 const KeyTypeED25519 = arieskms.ED25519
 
+// Store defines the storage capability for local KMS.
+type Store interface {
+	// Put stores the given key under the given keysetID.
+	Put(keysetID string, key []byte) error
+	// Get retrieves the key stored under the given keysetID. If no key is found, the returned error.
```
DRK3, line 25: `then an error is returned` or something similar

cmd/wallet-sdk-gomobile/localkms/localkms.go
```diff
@@ -58,3 +70,8 @@ func (k *KMS) ExportPubKey(keyID string) ([]byte, error) {
 func (k *KMS) GetSigningAlgorithm(keyID string) (string, error) {
 	return k.goAPILocalKMS.GetSigningAlgorithm(keyID)
 }
+
+// GetCrypto returns Crypto instance that can performs crypto ops with keys created by this kms.
```
DRK3, line 74: `perform` instead of `performs`

DRK3 (changes requested):
See my comment regarding the internal file

sudeshrshetty (approved):

talwinder50 (approved):

cmd/wallet-sdk-gomobile/credentialquery/credentialquery.go
vkubiv: if it is internal in the outside package we have circular dependencies. root package began to depend on cmd.
If make it internal inside cmd it will break the mobile bindings build.

cmd/utilities/go.mod
```diff
@@ -0,0 +1,48 @@
+// Copyright Avast Software. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+module github.com/trustbloc/wallet-sdk/cmd/utilities
```
DRK3, line 5: I think we should directly name this `gomobilewrappers` - it's a little bit more specific of a name

vkubiv: I named it "utilities"  because it is probable that we will have other utility functionality, not only gomobilewrappers

vkubiv: so we can put it there. Also putting gomobilewrappers into utilities also shows that it is not some independent functionality but just some utilities for wallet-sdk-gomobile

DRK3 (approved):
Left one more comment to consider regarding the new package name, but otherwise looks good

cmd/utilities/go.mod
DRK3: Seems OK to me in that case - the only other things I can think of are to have multiple go mod files (one for each type of utility), or, move them back into `wallet-sdk-gomobile` but then update the gomobile build scripts to ignore the wrappers folder... but as discussed that will take some time. Maybe in the future as the project grows we can see if a different folder structure (for everything) would make more sense

DRK3: I think it's OK to merge unless anyone else has any ideas

rolsonquadras: Probably it would be better to add a prefix which best describes the module to this rather having a generic name `utilities` @vkubiv. We can have a follow up PR.

PR #87, opened 2022-11-29
feat: Credential Preview screen

talwinder50:
This is the initial cut will refine more in the upcoming prs 
#86  Background color and logo need to be taken into consideration separately as colors data which we are expecting flutter is color widget are not rendering those values

IMG_0251

### DIFF
--- a/demo/app/lib/views/otp.dart
+++ b/demo/app/lib/views/otp.dart
@@ -54,8 +54,7 @@ class _OTPPage extends State<OTP> {
                       _fieldThree.text +
                       _fieldFour.text;
                 });
-                var requestCredentialResp = await WalletSDKPlugin.requestCredential(_otp!).toString();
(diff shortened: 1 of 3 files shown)


demo/app/lib/views/otp.dart
```diff
@@ -54,8 +54,8 @@ class _OTPPage extends State<OTP> {
                       _fieldThree.text +
                       _fieldFour.text;
                 });
-                var requestCredentialResp = await WalletSDKPlugin.requestCredential(_otp!).toString();
-                print(requestCredentialResp);
+                var requestCredentialResp = await WalletSDKPlugin.requestCredential(_otp!);
+                print("what is the data received $requestCredentialResp");
```
rolsonquadras, line 58: Is this checked in by mistake ?

vkubiv (approved):

PR #100, opened 2022-12-09
feat: Improve names used in mobile bindings

DRK3:
* Refactored wallet-sdk-gomobile to improve the generated names for the mobile bindings (reducing repeated words, enhancing clarity, etc).
* Added documentation on package/module names for the mobile bindings.
* Removed unimplemented stub types. These really only existed before so we could test out the gomobile tool. We can add them back in once we actually go to implement them.


cmd/wallet-sdk-gomobile/docs/packages.md
```diff
@@ -0,0 +1,400 @@
+# Packages
```
DRK3, line 1: I recommend viewing this file directly on my branch instead of here to see the rendered view. Note that you can use the rich diff option above, but it shows a horizontal scrollbar that won't actually be there

cmd/wallet-sdk-gomobile/docs/packages.md
```diff
@@ -0,0 +1,400 @@
...
+The left-most side in the tables shows Go source code snippets, rather than how they would look when imported in another
+file. The middle and right-most side show how the code looks when importing that same functionality in via the bindings.
+The reason for comparing between Go and the bindings like this is that the Go code is not intended to be imported
+directly - the mobile wrappers exist only to make the Go code compatible with the `gomobile` tool, so seeing how it
+would look imported is not so useful. However, the way you import from the generated mobile bindings is important,
+and so that's why we show that here. The Go source code is really just meant to help demonstrate the naming patterns
+to the reader (as the names used for packages/modules in the bindings are generated automatically by `gomobile` based
+on the Go names).
+
+### API
```
DRK3, line 72: In the future, we should consider finding a better way to organize the stuff in the `api` folder. I don't think it's very clear to a caller why certain things are in here, or why it's named `api`. Right now it's basically a "common" type of folder.

DRK3, line 72: I didn't make any changes to the `api` names.

cmd/wallet-sdk-gomobile/docs/packages.md
```diff
@@ -0,0 +1,400 @@
...
+import Walletsdk
+
+ApiCredentialReader
+```
+
+</td>
+</tr>
+</table>
+
+### Credential Querying
```
DRK3, line 114: The change I made here allowed me to eliminate most of the repetition in the bindings names. For example, for iOS, using the API from Swift used to look like this:

```
let query = CredentialqueryNewQuery(...)
query.query(...)
```

Also, it was somewhat confusing that the type for doing a query was itself called `Query`, which didn't quite make sense, at the type was really something that could do queries, not the query itself. I renamed it to `Inquirer`.

cmd/wallet-sdk-gomobile/docs/packages.md
```diff
@@ -0,0 +1,400 @@
...
+import Walletsdk
+
+CredentialNewInMemoryDB()
+```
+
+</td>
+</tr>
+</table>
+
+### Local KMS
```
DRK3, line 246: I had trouble with this one - I couldn't figure out a nice way to eliminate the work repetition here... let me know if you have any ideas

cmd/wallet-sdk-gomobile/docs/packages.md
```diff
@@ -0,0 +1,400 @@
...
+import Walletsdk
+
+LocalkmsNewKMS(...)
+```
+
+</td>
+</tr>
+</table>
+
+### Linked Domains
```
DRK3, line 285: Shortened names.

cmd/wallet-sdk-gomobile/docs/packages.md
```diff
@@ -0,0 +1,400 @@
...
+
+<table>
+<tr>
+<td> Go Source Code </td> <td> Java (as called from Kotlin) </td> <td> Obj-C (as called from Swift) </td>
+</tr>
+<tr>
+<td> 
+
+```
+package openid4ci
```
DRK3, line 334: I used to have the standalone credential display resolver in a separate package. I moved it into the same package as the interaction since it's specific to `openid4ci` anyway. It has the side benefit of resolving the word duplication issues and reduces the number of Android packages by 1.

rolsonquadras (approved):

cmd/wallet-sdk-gomobile/docs/packages.md
```diff
@@ -0,0 +1,400 @@
...
+
+DidNewCreatorWithKeyWriter(...)
+DidNewResolver()
+```
+
+</td>
+</tr>
+</table>
+
+### In-Memory Credential DB
```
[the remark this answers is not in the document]
DRK3: (in Aries, we had the provider vs store distinction because a provider could create multiple stores, however here we don't need the multiple level structure)

vkubiv (approved):

PR #108, opened 2022-12-13
docs: Mobile bindings usage

DRK3:

### DIFF
--- a/cmd/wallet-sdk-gomobile/docs/usage.md
+++ b/cmd/wallet-sdk-gomobile/docs/usage.md
@@ -1,0 +1,401 @@
+# SDK Usage
+
+Last updated: December 12, 2022 (commit `88e63eaac7e66835e779cb0a718337505bf05dcd`)
+
+This guide explains how to use this SDK in Android or iOS code.
+
+For the sake of readability, the following is omitted in the code examples:
+* Error handling
+* Optionals unwrapping (in Swift examples)
+
+## API Package
+
+The API package contains various interfaces and types that are used across multiple places in the SDK.
+Wallet-SDK contains implementations of these various interfaces that can be used to get an application up and running
+fairly quickly.
+Implementations of these interfaces can be done in on the mobile side and injected in to the API methods that use them.
+A few examples of when you might want to do this:
+* Implementing platform-specific credential storage (perhaps in a secure enclave)
+* Implementing platform-specific crypto functionality (perhaps leveraging device-specific security chips)
+
+## In-Memory Credential Storage
+The credential package contains an in-memory credential storage implementation that can be used to satisfy both
+credential reader and credential writer interfaces. As it only uses in-memory storage, you will probably want to
+create your own implementation in your mobile code that uses platform-specific storage.
+
+### Examples
+
+#### Kotlin (Android)
+
+```kotlin
+import dev.trustbloc.wallet.sdk.api.*
+import dev.trustbloc.wallet.sdk.credential.*
+
+val db = credential.newInMemoryDB()
+
+val vc = api.JSONObject()
+vc.setData = "VC JSON goes here"
+
+db.add(vc)
+
+retrievedVC = db.get("VC_ID")
+
+retrievedVCs = db.getAll()
+
+db.remove("VC_ID")
+```
+
+#### Swift (iOS)
+
+```swift
+import Walletsdk
+
+let db = CredentialNewInMemoryDB()
+
+let vc = ApiJSONObject(data: "VC JSON goes here")
+
+db.add(vc)
+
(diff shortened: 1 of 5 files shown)


cmd/wallet-sdk-gomobile/docs/usage.md
```diff
@@ -0,0 +1,358 @@
...
+If the store argument in the constructor is left null/nil, then an in-memory key store will be used.
+To use your own key store implementation, you can implement a store on the mobile side and inject it in to the
+constructor.
+
+### Examples
+
+#### Kotlin (Android)
+
+```kotlin
+import dev.trustbloc.wallet.sdk.localkms
```
DRK3, line 77: I would like to suggest the following changes to our KMS implementation:

### Rename `localkms` to `tinkkms`

In reality, there's nothing "local" about the KMS. You can inject in any store you want. The only real defining characteristic that I can see if that it uses Google's Tink crypto library. Therefore, I think `tinkkms` would be a better name.

### Remove the implicit behaviour of null argument -> use in-memory storage

I think it would be cleaner and clearer if we make the store a required argument, and if the caller wants to use an in-memory store, then they should explicitly pass one in (we will provide this implementation so that the caller can use it, of course). If the caller passes null, we throw an error.

Thoughts?

Moopli: > In reality, there's nothing "local" about the KMS.
I guess the distinction is that this KMS implementation will have private keys intermittently present in-memory on the client device - even if it uses a remote store - because it loads key material, unwraps it with a local in-memory master key, and uses the key material locally for crypto operations.

To contrast, remote KMS never holds private keys on the client device.

rolsonquadras:
@DRK3 Can we add the link to this page in main README?

DRK3:
> @DRK3 Can we add the link to this page in main README?

Done!

cmd/wallet-sdk-gomobile/docs/usage.md
DRK3: Good point @Moopli! I think this is an important thing for consumers of this API to be aware of, so I've updated the documentation accordingly. Let me know if you have any feedback on this.

rolsonquadras (approved):

talwinder50 (approved):

cmd/wallet-sdk-gomobile/docs/usage.md
```diff
@@ -0,0 +1,401 @@
...
+
+let createDIDOpts = ApiCreateDIDOpts(keyID: keyHandle.keyID, verificationType: DidEd25519VerificationKey2018)
+
+let didDocResolution = didCreator.create("key", createDIDOpts) // Create a did:key doc
+```
+
+## OpenID4CI
+
+The OpenID4CI package contains an API that can be used by a [holder](https://www.w3.org/TR/vc-data-model/#dfn-holders)
+to go through the [OpenID4CI](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html) flow.
```
sudeshrshetty, line 241: change links to spec according to this https://github.com/trustbloc/vcs/tree/main/component/wallet-cli#oidc4cioidc4vc-flow

Links you used keeps changing every month, we can not always keep up with those updates.

Better we mention exact version we choose to implement.

cmd/wallet-sdk-gomobile/docs/usage.md
```diff
@@ -0,0 +1,401 @@
...
+let userPIN = "1234"
+let requestCredentialOpts = Openid4ciNewCredentialRequestOpts(userPIN)
+let credentials = interaction.requestCredential(requestCredentialOpts) // Should probably store these somewhere
+let displayData = interaction.resolveDisplay("en-US") // Optional (but useful)
+```
+
+## OpenID4VP
+
+The OpenID4VP package contains an API that can be used by a [holder](https://www.w3.org/TR/vc-data-model/#dfn-holders)
+to go through the [OpenID4VP](https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0-07.html) flow.
```
sudeshrshetty, line 335: wrong version --> https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0-ID1.html

cmd/wallet-sdk-gomobile/docs/usage.md
DRK3: Done!

cmd/wallet-sdk-gomobile/docs/usage.md
DRK3: Done!

sudeshrshetty (approved):

PR #120, opened 2022-12-17
feat: DID linked domains service validation function

DRK3:


vkubiv (approved):

sandrask (approved):

cmd/wallet-sdk-gomobile/did/wellknown.go
```diff
@@ -0,0 +1,49 @@
...
+type ValidationResult struct {
+	IsValid    bool
+	ServiceURL string
+}
+
+// ValidateService validates the given DID's service against its well-known DID configuration.
+// It returns a ValidationResult.
+// The DID document must specify only a single service. If there are multiple URLs for a given service, only the
+// first will be checked.
+func ValidateService(did string, resolver api.DIDResolver) (*ValidationResult, error) {
```
rolsonquadras, line 32: Should this be names ValidateWellKnownDIDConfig as that's the actual validation? There can be multiple services, but this specific validation is for Well Known DID Config.

DRK3: I think `ValidateWellKnownDIDConfig` would be somewhat misleading, as the main point of this function isn't validating the well-known DID config itself, but rather is validating the service in the DID doc _against_ its well-known DID config. (The function in Aries that I call does happen to also validate the well-known DID config along the way, but I think that's just a piece of the overall picture)

rolsonquadras: There could be multiple services inside a DID Document. We are validating services with `"type": "LinkedDomains"` - https://identity.foundation/.well-known/resources/did-configuration/#example-7.

sudeshrshetty: agree with @rolsonquadras 
`VaidateService` is too generic name, should be called `ValidateLinkedDomain, ValidateWellKnown or ValidateWellKnownConfig`

DRK3: `ValidateLinkedDomains` sounds like a better name to me. I've renamed it to that (and also added in a check to ensure the service is of type `LinkedDomains`)

pkg/did/wellknown/wellknown.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	didDocResolution, err := resolver.Resolve(did)
+	if err != nil {
+		return false, "", fmt.Errorf("failed to resolve DID: %w", err)
+	}
+
+	if len(didDocResolution.DIDDocument.Service) == 0 {
+		return false, "", errors.New("resolved DID document has no services specified")
+	} else if len(didDocResolution.DIDDocument.Service) > 1 {
```
rolsonquadras, line 48: This condition is not valid. The DID Document can have more than 1 services. For LinkedDomain, we said we will restrict to 1. If DID Document has 1 Linked Domain type and a different type service, this case will fail.

DRK3: Fixed.

rolsonquadras (approved):

sudeshrshetty (approved):

PR #151, opened 2023-01-06
feat(sdk,app): Mobile VC refactor + IssuerID method

DRK3:
* Refactored the gomobile VC type to hold the Aries VC type internally instead of the marshalled VC. This is more efficient since it eliminates extra credential reparsing operations.
* Added a serialize method to the new VC type.
* Added a parse method so a caller can parse a serialized VC into the gomobile VC type.
* Added a method to the VerifiableCredential mobile type that can be used to get the issuer ID (usually a DID).
* Switched the database methods to the VerifiableCredential type from the previous generic JSON type.


cmd/wallet-sdk-gomobile/api/credentials.go
```diff
@@ -6,18 +6,47 @@ SPDX-License-Identifier: Apache-2.0
...
+	if !ok {
+		return "", errors.New("unexpected VC subject type")
+	}
+
+	if len(subjects) == 0 {
+		return "", errors.New("no subjects in VC")
+	}
+
+	return subjects[0].ID, nil
+}
```
DRK3, line 48: While this gomobile-compatible Verifiable Credential + `SubjectID` method implementation is functional, it has some issues. The biggest problem is that since it internally stores the VC in marshalled form (see line 19), it has to get reparsed for various operations (ex. the new SubjectID method I added in this PR). This can be slow, especially with the network loader (which I'm using above).

I see a few potential solutions, but each has some drawbacks so I wanted to gather some feedback first.

Solution 1: Change the internal representation to be the parsed VC type from Aries (line 19). I can provide a method on the Verifiable Credential wrapper to return a marshalled version that the caller can use if they need it (note: in many places the SDK user does not need the full "raw" marshalled VC anyway). The advantage here is that we can eliminate the constant reparsing and remarshalling we have in a lot of places. Instead, the parsing will only happen once in the methods that produce the VCs in the first place, and those places already have (or will have) injectable document loaders to make parsing faster.

The one drawback with this method is that this will result in a "skipped field/skipped method" type of line in the binding headers, since that internal VC representation (line 19) has to be public (or have a public setter type method) since the various SDK packages need to be able to set it. Thus far, in developing the SDK I've been trying to avoid this scenario since I felt it was a bit improper to have any "skipped" field sort of messages in the headers, but I don't think this would actually cause any issues for the SDK users (only perhaps looking a bit weird if they manually look into the headers) as they should still be able to use those VC wrapper objects. We don't actually need/want them to be able to set that field anyway.

I think this is probably the best solution.

Solution #2: Keep it as-is and make the reparsing faster by allowing a document loader to be injected in to the SubjectID method (or finding a way to allow the VC wrapper to get the document loader from the source where it was originally generated - probably difficult/ugly) or have some sort of global static document loader. This won't address all the issues of the reparsing (it'll still happen), but perhaps it'll be fast enough to not be a concern.

Let me know what you think!

cmd/wallet-sdk-gomobile/api/credentials.go
```diff
@@ -6,18 +6,47 @@ SPDX-License-Identifier: Apache-2.0
...
 
 // NewVerifiableCredential creates a new VerifiableCredential.
-func NewVerifiableCredential(content string) *VerifiableCredential {
+func NewVerifiableCredential(content []byte) *VerifiableCredential {
 	return &VerifiableCredential{
 		Content: content,
 	}
 }
 
+// SubjectID returns the ID of the first subject identified in the VC.
```
sudeshrshetty, line 29: what you mean by first subject? there is only one subject per credential

cmd/wallet-sdk-gomobile/api/credentials.go
DRK3: (technically a third solution, but problematic: have everything that needs the VC type in the same module, and that way that internal VC representation can be private, avoiding the warning)

DRK3: I've implemented it per our discussion (holding aries VC type internally).

cmd/wallet-sdk-gomobile/api/vcparse/vcparse.go
```diff
@@ -0,0 +1,54 @@
+/*
```
DRK3, line 1: I wanted to have this function in the same place as the VC definition, but this wasn't possible due to a circular import issue. I can think of a way around it, but it would result in a lot of code duplication.

vkubiv (approved):

cmd/wallet-sdk-gomobile/wrapper/credentialreader.go
```diff
@@ -26,45 +23,25 @@ type CredentialReaderWrapper struct {
...
 // Get wraps Get of api.CredentialReader.
 func (r *CredentialReaderWrapper) Get(id string) (*verifiable.Credential, error) {
-	content, err := r.CredentialReader.Get(id)
+	vc, err := r.CredentialReader.Get(id)
 	if err != nil {
 		return nil, err
 	}
 
-	cred, err := verifiable.ParseCredential(content.Data, verifiable.WithDisabledProofCheck(),
-		verifiable.WithJSONLDDocumentLoader(r.DocumentLoader))
```
vkubiv, line 35: if DocumentLoader no used any where else in CredentialReaderWrapper we can delete it

vkubiv (approved):

sudeshrshetty (approved):

cmd/wallet-sdk-gomobile/wrapper/credentialreader.go
DRK3: Good catch! I'll make a followup PR shortly.

PR #168, opened 2023-01-10
fix: Circular dependency

DRK3:
The cmd/wallet-sdk-gomobile module was depending on the cmd/utilities module (with a replace statement), and the cmd/utilities module was depending on cmd/wallet-sdk-gomobile module (with a recent commit tag). This causes import issues when updates are made to the interfaces in cmd/wallet-sdk-gomobile.

This commit fixes that by moving the code from cmd/utilities to cmd/wallet-sdk-gomobile and making it a package instead of a module. In order to avoid the exported types and methods from appearing in the bindings (as they are internal details), I modified the build scripts to exclude that package.

This solution allows us to reuse those wrapper objects while preventing a circular dependency and also prevents them from showing up in the bindings.


DRK3:
Note that gomobile doesn't allow me to use an `internal` go package (which would have also solved the problem without requiring the script changes).

Moopli (approved):

cmd/wallet-sdk-gomobile/scripts/build_android_bindings.sh
```diff
@@ -0,0 +1,7 @@
+# This script needs to be called from the same folder that the wallet-sdk-gomobile makefile is in.
```
rolsonquadras, line 1: Can we add licence header to these new files ?

rolsonquadras (approved):

cmd/wallet-sdk-gomobile/scripts/build_android_bindings.sh
DRK3: Done!

PR #170, opened 2023-01-11
feat(sdk): error model.

vkubiv:

### DIFF
--- a/cmd/wallet-sdk-gomobile/openid4vp/errors.go
+++ b/cmd/wallet-sdk-gomobile/openid4vp/errors.go
@@ -1,0 +1,20 @@
+/*
+Copyright Avast Software. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package openid4vp
+
+// Constants' names and reasons are obvious so they do not require additional comments.
+// nolint:golint,nolintlint
+const (
+	module                     = "OVP"
+	CredentialParseFailedError = "CREDENTIAL_PARSE_FAILED" //nolint:gosec //false positive
+)
+
+// Constants' names and reasons are obvious so they do not require additional comments.
+// nolint:golint,nolintlint
+const (
+	CredentialParseFailedCode = iota + 1000
+)
--- a/cmd/wallet-sdk-gomobile/walleterror/error.go
+++ b/cmd/wallet-sdk-gomobile/walleterror/error.go
@@ -1,0 +1,71 @@
+/*
+Copyright Avast Software. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package walleterror defines the error model for the Go mobile API.
+package walleterror
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/trustbloc/wallet-sdk/pkg/walleterror"
+)
+
+// Error represents error returned by go mobile api.
+type Error struct {
+	Code     string `json:"code"`
+	Category string `json:"category"`
+	Details  string `json:"details"`
+}
+
+// ToMobileError translates go api errors to go mobile errors.
+func ToMobileError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	var result *Error
+
+	var walletError *walleterror.Error
+
+	if errors.As(err, &walletError) {
+		result = &Error{
+			Code:     walletError.Code,
+			Category: walletError.Scenario,
+			Details:  walletError.ParentError,
+		}
+	} else {
+		result = &Error{
+			Code:     "UKN2-000",
+			Category: "UNEXPECTED_ERROR",
+			Details:  err.Error(),
+		}
+	}
+
+	formatted, fmtErr := json.Marshal(result)
+	if fmtErr != nil {
+		return fmt.Errorf("unexpected format error: %w", fmtErr)
+	}
+
+	return errors.New(string(formatted))
+}
+
+// Parse used to parse exception message on mobile side.
+func Parse(message string) *Error {
+	walletErr := &Error{}
+
+	err := json.Unmarshal([]byte(message), walletErr)
+	if err != nil {
+		return &Error{
+			Code:     "GNR2-000",
+			Category: "GENERAL_ERROR",
+			Details:  message,
+		}
+	}
+
+	return walletErr
+}
--- a/demo/app/android/app/src/main/kotlin/walletsdk/flutter/app/MainActivity.kt
+++ b/demo/app/android/app/src/main/kotlin/walletsdk/flutter/app/MainActivity.kt
@@ -8,6 +8,7 @@ import io.flutter.plugin.common.MethodChannel
 import dev.trustbloc.wallet.sdk.localkms.KMS
 import dev.trustbloc.wallet.sdk.did.Resolver
 import dev.trustbloc.wallet.sdk.ld.DocLoader
+import dev.trustbloc.wallet.sdk.walleterror.Walleterror
 import dev.trustbloc.wallet.sdk.localkms.Localkms
 import dev.trustbloc.wallet.sdk.localkms.SignerCreator
 import io.flutter.plugin.common.MethodCall
@@ -57,10 +58,16 @@ class MainActivity : FlutterActivity() {
                                 result.success(userPinRequired)
 
                             } catch (e: Exception) {
+                                val err = Walleterror.parse(e.message)
+                                // Add custom error handling logic here basing on code and error properties
+                                println("code: ${err.code}")
+                                println("error: ${err.category}")
+                                println("details: ${err.details}")
+
                                 result.error(
                                         "Exception",
                                         "Error while authorizing the oidc vc flow",
-                                        e
+                                        "code: ${err.code}, error: ${err.category}, details: ${err.details}"
                                 )
 
                             }
@@ -187,9 +194,9 @@ class MainActivity : FlutterActivity() {
         return resp
     }
 
-    public fun resolveCredentialDisplay(call: MethodCall) : String?{
+    public fun resolveCredentialDisplay(call: MethodCall): String? {
         val openID4CI = this.openID4CI
-            ?: throw java.lang.Exception("openID4CI not initiated. Call authorize before this.")
+                ?: throw java.lang.Exception("openID4CI not initiated. Call authorize before this.")
 
         val resp = openID4CI.resolveCredentialDisplay() ?: return null
 
@@ -197,7 +204,7 @@ class MainActivity : FlutterActivity() {
     }
 
 
-       private fun createDID(call: MethodCall): String {
+    private fun createDID(call: MethodCall): String {
         val kms = this.kms ?: throw java.lang.Exception("SDK is not initialized, call initSDK()")
 
         val creatorDID = Creator(kms as KeyWriter)
--- a/internal/testutil/errors.go
+++ b/internal/testutil/errors.go
@@ -1,0 +1,20 @@
+/*
+Copyright Avast Software. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package testutil
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// RequireErrorContains custom require function to check that error contains a substring.
+// nolint:thelper,nolintlint //it isn't helper function it is require function.
+func RequireErrorContains(t *testing.T, err error, errString string) {
+	require.Error(t, err)
+	require.Contains(t, err.Error(), errString)
+}
--- a/pkg/credentialquery/credentialquery.go
+++ b/pkg/credentialquery/credentialquery.go
@@ -8,6 +8,7 @@ SPDX-License-Identifier: Apache-2.0
 package credentialquery
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/hyperledger/aries-framework-go/pkg/doc/presexch"
@@ -15,6 +16,7 @@ import (
 	"github.com/piprate/json-gold/ld"
 
 	"github.com/trustbloc/wallet-sdk/pkg/api"
+	"github.com/trustbloc/wallet-sdk/pkg/walleterror"
 )
 
 // Instance implements querying credentials using presentation definition.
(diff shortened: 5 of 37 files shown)


cmd/wallet-sdk-gomobile/walleterror/error.go
```diff
@@ -0,0 +1,55 @@
...
+
+	if errors.As(err, &walletError) {
+		result = &Error{
+			Code:    walletError.Code,
+			Error:   walletError.ErrorName,
+			Details: walletError.Details,
+		}
+	} else {
+		result = &Error{
+			Code:    "UKN2-000",
```
sudeshrshetty, line 30: values for these codes has to be supplied by caller.

pkg/openid4ci/errors.go
```diff
@@ -0,0 +1,23 @@
...
+	ClientConfigNoClientIDProvidedError = "CLIENT_CONFIG_NO_CLIENT_ID_PROVIDED"
+
+	ClientConfigNoSignerProviderProvidedCode  = module + "0-0004"
+	ClientConfigNoSignerProviderProvidedError = "CLIENT_CONFIG_NO_SIGNER_PROVIDER_PROVIDED"
+
+	ClientConfigNoDIDResolverProvidedCode  = module + "0-0005"
+	ClientConfigNoDIDResolverProvidedError = "CLIENT_CONFIG_DID_RESOLVER_PROVIDED"
+
+	PreAuthorizedCodeRequiredCode  = module + "0-0001"
+	PreAuthorizedCodeRequiredError = "PRE_AUTHORIZED_CODE_REQUIRED"
```
sudeshrshetty, line 22: Why cant we use `iota`, and let `walleterror` build the whole error code based on scenario.

Example: https://github.com/hyperledger/aries-framework-go/blob/main/pkg/controller/command/vcwallet/command.go#L269

demo/app/android/app/src/main/kotlin/walletsdk/flutter/app/MainActivity.kt
```diff
@@ -57,10 +58,16 @@ class MainActivity : FlutterActivity() {
                                 result.success(userPinRequired)
 
                             } catch (e: Exception) {
+                                val err = Walleterror.parse(e.message)
```
DRK3, line 61: Is it possible to pass a struct through to the exception instead of having to parse the message string? (no idea if this is possible - just wanted to check and see if gomobile/the bindings allow it)

pkg/credentialquery/credentialquery.go
```diff
@@ -65,14 +67,23 @@ func (c *Instance) Query(
 	credentials := qOpts.credentials
 	if len(credentials) == 0 {
 		if qOpts.credentialReader == nil {
-			return nil, fmt.Errorf("credentials array or credential reader option should be set")
+			return nil, walleterror.NewValidationError(
+				module,
+				CredentialReaderNotSetCode,
+				CredentialReaderNotSetError,
+				fmt.Errorf("credentials array or credential reader option should be set"))
```
DRK3, line 74: Should be `must be set` rather than `should be set`

pkg/credentialquery/errors.go
```diff
@@ -0,0 +1,24 @@
...
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package credentialquery
+
+// nolint: golint,nolintlint
+const (
+	module                               = "CRQ"
+	CredentialReaderNotSetError          = "CREDENTIAL_READER_NOT_SET"     //nolint: gosec
+	CredentialReaderReadFailedError      = "CREDENTIAL_READER_READ_FAILED" //nolint:gosec
```
DRK3, line 13: Reason for the `nolint: golint,nolintlint`? I'm guessing you did that because the linter requires a comment for each of those exported constants, and their purposes are pretty obvious already, but it's a good idea to at least add the comment explaining the reason for the nolint. Same thing for the `gosec` - in this case, it's a false positive as the linter flags anything with the word `credential` in it, but I still think it's a good idea to at least add `false positive` in the nolint comment to explain it, since it can look concerning to see a `gosec` warning being bypassed without an explanation. Also, one more minor thing - the spacing between the `nolint` and `gosec` should be consistent.

DRK3: (for consistency with the rest of the codebase, I would suggest just removing the `nolint: golint,nolintlint` and adding the export comments, even if it seems somewhat redundant)

pkg/credentialquery/errors.go
```diff
@@ -0,0 +1,24 @@
...
+	CreateVPFailedError                  = "CREATE_VP_FAILED"
+	NoCredentialSatisfyRequirementsError = "NO_CREDENTIAL_SATISFY_REQUIREMENTS" //nolint:gosec
+)
+
+// nolint: golint,nolintlint
```
DRK3, line 18: Same note here regarding the nolint

pkg/did/resolver/resolver_test.go
```diff
@@ -68,6 +69,11 @@ func TestDIDResolver_InvalidDID(t *testing.T) {
 
 	didDocResolution, err := didResolver.Resolve("did:example:abc")
 	require.Error(t, err)
-	require.EqualError(t, err, "resolve did:example:abc : did method example not supported for vdr")
+	requireErrorContains(t, err, "resolve did:example:abc : did method example not supported for vdr")
 	require.Empty(t, didDocResolution)
 }
+
+func requireErrorContains(t *testing.T, err error, errString string) { //nolint:thelper
```
DRK3, line 76: Reason for nolint?

pkg/openid4ci/errors.go
```diff
@@ -0,0 +1,44 @@
+/*
+Copyright Avast Software. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package openid4ci
+
+// nolint: golint,nolintlint
```
DRK3, line 9: Same note about the nolints

pkg/openid4ci/openid4ci.go
```diff
@@ -97,14 +111,22 @@ func (i *Interaction) Authorize() (*AuthorizeResult, error) {
 // Relevant sections of the spec:
 // https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-7
 // https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-8
-func (i *Interaction) RequestCredential(credentialRequestOpts *CredentialRequestOpts) ([]CredentialResponse, error) {
+func (i *Interaction) RequestCredential(credentialRequestOpts *CredentialRequestOpts) ([]CredentialResponse, error) { //nolint:funlen,lll
```
DRK3, line 114: Reason for the nolint?

pkg/walleterror/walleterror.go
```diff
@@ -0,0 +1,59 @@
+/*
+Copyright Avast Software. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package walleterror define error model for go api.
```
DRK3, line 7: Should say `Package walleterror defines the error model for the Go api.`

cmd/wallet-sdk-gomobile/walleterror/error.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Error represents error returned by go mobile api.
+type Error struct {
+	Code    string `json:"code"`
+	Error   string `json:"error"`
+	Details string `json:"details"`
+}
+
+// Translate translates go api errors to go mobile errors.
+func Translate(err error) error {
```
DRK3, line 26: I would suggest renaming this function to `ToMobileError` for a few reasons:

1. In the context of error messages (especially ones that may get displayed on an end-user's screen), `Translate` can be confused with language translation/localization.
2. `ToMobileError` is more precise anyway.

cmd/wallet-sdk-gomobile/walleterror/error.go
```diff
@@ -0,0 +1,71 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/trustbloc/wallet-sdk/pkg/walleterror"
+)
+
+// Error represents error returned by go mobile api.
+type Error struct {
+	Code    string `json:"code"`
+	Error   string `json:"error"`
```
DRK3, line 21: It's not clear from looking at this why the `Error` struct also has a field named `Error`. What is `Error.Error` supposed to be? Is it more like an error type/category? If so, I would suggest `Type` or `Category` (`Type` might be a reserved word in the bindings?)

cmd/wallet-sdk-gomobile/walleterror/error.go
```diff
@@ -0,0 +1,71 @@
+/*
+Copyright Avast Software. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package walleterror define error model for go mobile api.
```
DRK3, line 7: `Package walleterror defines the error model for the Go mobile API.`

cmd/wallet-sdk-gomobile/openid4vp/errors.go
```diff
@@ -0,0 +1,18 @@
+/*
+Copyright Avast Software. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package openid4vp
+
+// nolint: golint,nolintlint
```
DRK3, line 9: Same note about the nolint as in other places

pkg/openid4ci/openid4ci_test.go
```diff
@@ -651,3 +651,8 @@ func makeMockDoc(keyManager *arieslocalkms.LocalKMS) (*did.Doc, error) {
 
 	return newDoc, nil
 }
+
+func requireErrorContains(t *testing.T, err error, errString string) { //nolint:thelper
```
DRK3, line 655: Reason for the nolint

demo/app/android/app/src/main/kotlin/walletsdk/flutter/app/MainActivity.kt
vkubiv: I don't find way.

pkg/credentialquery/errors.go
vkubiv: I think we should create a separate task for this. Comments should be meaningful, contains reasons why that error can happen, and if it is caused by wrong use of API, the comment should suggest a solution, e.g. "Please create DID before calling this function" etc. But this is very time-consuming, and I think should be done module by module.

pkg/openid4ci/openid4ci.go
vkubiv: function becomes longer as we have multiline errors, and goes beyond line limits.  and I don't want to mess lint comment with the function comment so I added it to the end of the line so it required 'lll'. I don't have a good solution here.

pkg/credentialquery/errors.go
vkubiv: done

pkg/did/resolver/resolver_test.go
vkubiv: done

pkg/openid4ci/errors.go
vkubiv: done

pkg/walleterror/walleterror.go
```diff
@@ -0,0 +1,59 @@
...
+
+import "fmt"
+
+const (
+	validationError = 0
+	executionError  = 1
+	systemError     = 2
+)
+
+// Error represents error returned by go api.
```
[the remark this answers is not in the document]
vkubiv: done

pkg/walleterror/walleterror.go
vkubiv: done

pkg/openid4ci/openid4ci.go
vkubiv: done

pkg/walleterror/walleterror.go
vkubiv: done

cmd/wallet-sdk-gomobile/walleterror/error.go
vkubiv: done

cmd/wallet-sdk-gomobile/walleterror/error.go
vkubiv: done, now named Category

cmd/wallet-sdk-gomobile/walleterror/error.go
vkubiv: done

pkg/openid4ci/openid4ci_test.go
vkubiv: done

pkg/walleterror/walleterror.go
```diff
@@ -0,0 +1,59 @@
...
+func NewValidationError(module string, code int, errorName string, parentError error) *Error {
+	return &Error{
+		Code:        getErrorCode(module, validationError, code),
+		ErrorName:   errorName,
+		ParentError: parentError.Error(),
+	}
+}
+
+// NewExecutionError creates execution error.
+func NewExecutionError(module string, code int, errorName string, parentError error) *Error {
```
sudeshrshetty, line 35: suggestion: instead of naming it parent error, we can call it `cause`

sudeshrshetty: `errorName` can be `scenario`

sudeshrshetty (approved):

pkg/credentialquery/errors.go
DRK3: That's fair

internal/testutil/errors.go
```diff
@@ -0,0 +1,20 @@
...
+package testutil
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// RequireErrorContains error helper testing function.
+// nolint:thelper,nolintlint //it isn't helper function it is require function.
```
DRK3, line 16: While I understand the reason for the `thelper` nolint, having line 15 say that this is a helper testing function and then immediately after on line 16 say that it isn't a helper function is pretty contradictory. I would suggest changing line 15 to just briefly describe what the function does to get around it.

pkg/openid4ci/openid4ci.go
DRK3: I see. I think we might end up with many functions requiring `nolint:funlen` in that case. I don't really have an immediate solution either. If this becomes common, we might consider disabling funlen repo-wide @rolsonquadras

pkg/openid4ci/errors.go
```diff
@@ -0,0 +1,46 @@
...
+const (
+	NoClientConfigProvidedCode = iota
+	ClientConfigNoUserDidProvidedCode
+	ClientConfigNoClientIDProvidedCode
+	ClientConfigNoSignerProviderProvidedCode
+	ClientConfigNoDIDResolverProvidedCode
+	PreAuthorizedCodeRequiredCode
+	InvalidIssuanceURICode
+	PinCodeRequiredCode
+	MetaDataFetchFailedCode
```
DRK3, line 40: "metadata" is one word, so this should be `MetadataFetchFailedCode`

DRK3 (approved):
Left minor comments only, great job!

PR #204, opened 2023-01-23
feat(sdk): Logger for OpenID4CI and OpenID4VP

DRK3:

### DIFF
--- a/cmd/wallet-sdk-gomobile/openid4ci/resolveopts.go
+++ b/cmd/wallet-sdk-gomobile/openid4ci/resolveopts.go
@@ -24,7 +24,7 @@ import (
 // We don't check here to see if multiple conflicting options are used - that's left up to the
 // goapicredentialschema.Resolve method that get called after this function returns.
 func prepareOpts(credentials *Credentials, issuerMetadata *IssuerMetadata,
-	preferredLocale string,
+	preferredLocale string, logger api.Logger,
 ) ([]goapicredentialschema.ResolveOpt, error) {
 	if credentials == nil {
 		return nil, errors.New("no credentials specified")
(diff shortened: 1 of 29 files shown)


cmd/wallet-sdk-gomobile/openid4ci/resolveopts.go
```diff
@@ -24,7 +24,7 @@ import (
 // We don't check here to see if multiple conflicting options are used - that's left up to the
 // goapicredentialschema.Resolve method that get called after this function returns.
 func prepareOpts(credentials *Credentials, issuerMetadata *IssuerMetadata,
-	preferredLocale string,
+	preferredLocale string, logger api.Logger,
```
sudeshrshetty, line 27: no other better way to provide logger instead of injecting it manually to each and every API?

pkg/credentialschema/credentialschema.go
```diff
@@ -10,11 +10,13 @@ SPDX-License-Identifier: Apache-2.0
 package credentialschema
 
 // Resolve resolves display information for some issued credentials based on an issuer's metadata.
+// operationType is the operation type to be used in log entries. It should reflect the context of where this function
```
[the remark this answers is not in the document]
DRK3: This isn't exposed to the mobile API.

cmd/wallet-sdk-gomobile/openid4ci/resolveopts.go
DRK3: I could make it a global variable that the caller can set in one place. However, I'm not entirely sure how the lifetime of that global variable will work when used in bindings (i.e. I'm concerned that it might become "unset" if certain things happen in an app using the SDK, like suspending it, etc)

DRK3: Also, the one you commented on is an internal function - the caller only has to pass in a logger at the beginning of the OpenID4CI and OpenID4VP interactions - (or, if they want to call the Resolve function outside of the flow)

pkg/consolelogger/consolelogger.go
```diff
@@ -0,0 +1,34 @@
+/*
+Copyright Avast Software. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package consolelogger contains a simple logger implementation that writes log entries directly to the console.
+package consolelogger
```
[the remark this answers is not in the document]
DRK3: Done

DRK3:
Turns out this isn't quite what we need. Since I need to make some major changes, I'm going to reopen a new PR (with a new commit message + branch)

PR #212, opened 2023-01-24
feat(sdk,app): Smaller mobile bindings

DRK3:
* Decreased the size of the Android bindings by ~44% (45.3 MB -> 25.5 MB) and decreased the size of the iOS bindings by ~47% (37.5 MB -> 20.1 MB) by adding linker flags that remove unneeded debug information from the generated binaries.

### DIFF
--- a/cmd/wallet-sdk-gomobile/scripts/build_android_bindings.sh
+++ b/cmd/wallet-sdk-gomobile/scripts/build_android_bindings.sh
@@ -11,4 +11,4 @@ packages_for_bindings=$(. scripts/generate_package_list.sh)
 
 android_java_pkg="dev.trustbloc.wallet.sdk"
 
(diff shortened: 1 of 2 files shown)


rolsonquadras:
> * Decreased the size of the Android bindings by ~44% and decreased the size of the iOS bindings by ~47% by adding linker flags that remove unneeded debug information from the generated binaries.

@DRK3 can you please add the absolute numbers as well ?

DRK3:
> > * Decreased the size of the Android bindings by ~44% and decreased the size of the iOS bindings by ~47% by adding linker flags that remove unneeded debug information from the generated binaries.
> 
> @DRK3 can you please add the absolute numbers as well ?

Ah, I just updated the commit message with the absolute file size at almost the exact same time as you made your comment :)

DRK3:
I reverted the change to the iOS bindings targets as is turns out that we do indeed need the `x86_64-simulator` bindings,

rolsonquadras (approved):

vkubiv (approved):

PR #215, opened 2023-01-26
feat(sdk): Activity log

DRK3:

### DIFF
--- a/cmd/wallet-sdk-gomobile/activitylogger/mem/mem.go
+++ b/cmd/wallet-sdk-gomobile/activitylogger/mem/mem.go
@@ -1,0 +1,56 @@
+/*
+Copyright Avast Software. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package mem contains a simple in-memory api.ActivityLogger implementation.
+package mem
+
+import (
+	"sync"
+
+	"github.com/trustbloc/wallet-sdk/cmd/wallet-sdk-gomobile/api"
+)
+
+// ActivityLogger is a simple in-memory api.ActivityLogger implementation.
+// New activities are appended to the array in the order that they are received.
+type ActivityLogger struct {
+	activities []*api.Activity
+	lock       sync.RWMutex
+}
+
+// NewActivityLogger returns a new ActivityLogger.
+func NewActivityLogger() *ActivityLogger {
+	return &ActivityLogger{}
+}
+
+// Log logs a single activity.
+// The activity is appended to the end of an internal array of activities.
+func (m *ActivityLogger) Log(activity *api.Activity) error {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	m.activities = append(m.activities, activity)
+
+	return nil
+}
+
+// Length returns the number of activities that this activity logger has logged.
+func (m *ActivityLogger) Length() int {
+	return len(m.activities)
+}
+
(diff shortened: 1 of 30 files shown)


rolsonquadras (approved):

cmd/wallet-sdk-gomobile/activitylogger/mem/mem.go
```diff
@@ -0,0 +1,56 @@
...
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	m.activities = append(m.activities, activity)
+
+	return nil
+}
+
+// Length returns the number of activities that this activity logger has logged.
+func (m *ActivityLogger) Length() int {
```
sudeshrshetty, line 40: What is the use of this function, underlying activity logger might just care about pushing this data model to store,

cmd/wallet-sdk-gomobile/activitylogger/mem/mem.go
```diff
@@ -0,0 +1,56 @@
...
+	return len(m.activities)
+}
+
+// AtIndex returns the activity stored at the given index.
+// Passing in 0 for the index will return the first activity in the underlying array. If this activity logger is only
+// being used in one thread at a time, then that activity will be first activity that this activity logger has logged,
+// and using higher indices will return more recent activities.
+// If the index passed in is out of bounds, then nil is returned.
+func (m *ActivityLogger) AtIndex(index int) *api.Activity {
```
sudeshrshetty, line 49: Activity logger interface is not a collection data model & it might not hold all the activity log entries, implementation can just push the data model to store and may not keep anything in memory.

vkubiv (approved):

sudeshrshetty (approved):

cmd/wallet-sdk-gomobile/activitylogger/mem/mem.go
DRK3: This isn't part of the interface - it's only in the in-memory implementation. It's intended for the reference demo app so we have some way to collect activities and show them in the UI.

cmd/wallet-sdk-gomobile/activitylogger/mem/mem.go
DRK3: Same as my previous comment - just for the reference app. The reference app needs some quick and easy way to read what activities took place.

cmd/wallet-sdk-gomobile/activitylogger/mem/mem.go
DRK3: (there is no underlying activity logger here)

cmd/wallet-sdk-gomobile/activitylogger/mem/mem.go
DRK3: (this is also not part of the interface)

cmd/wallet-sdk-gomobile/activitylogger/mem/mem.go
DRK3: I also make use of this type in the tests

cmd/wallet-sdk-gomobile/activitylogger/mem/mem.go
DRK3: Just to clarify - the interface is just a single method called "log" that allows for the data model to be pushed. Length() and AtIndex() are specific to my in-memory implementation

PR #218, opened 2023-01-27
fix(sdk): Runtime error when injecting local KMS store from mobile

DRK3:
Fixed an issue with the local KMS implementation that was causing a runtime error when using any store implementation that was injected in from mobile code. Also did some refactoring to bring the local KMS more in line with the other APIs.

### DIFF
--- a/cmd/wallet-sdk-gomobile/localkms/signercreator_test.go
+++ b/cmd/wallet-sdk-gomobile/localkms/signercreator_test.go
@@ -1,0 +1,115 @@
+/*
+Copyright Avast Software. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package localkms_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/btcsuite/btcutil/base58"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/jose/jwk/jwksupport"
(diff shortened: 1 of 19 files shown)


cmd/wallet-sdk-gomobile/localkms/signercreator_test.go
```diff
@@ -0,0 +1,115 @@
+/*
```
DRK3, line 1: I moved this from `localkms_test.go`

rolsonquadras (approved):

demo/app/ios/Runner/flutterPlugin.swift
```diff
@@ -68,7 +68,8 @@ public class SwiftWalletSDKPlugin: NSObject, FlutterPlugin {
     }
     
     private func initSDK(result: @escaping FlutterResult) {
-        kms = LocalkmsNewKMS(nil, nil)
+        let memKMSStore = LocalkmsNewMemKMSStore()
+        kms = LocalkmsNewKMS(memKMSStore, nil)
```
talwinder50, line 72: I will be changing to pass ios userdefault store right ??

talwinder50 (approved):

demo/app/ios/Runner/flutterPlugin.swift
DRK3: Yes :) (your implementation of that interface)

PR #222, opened 2023-02-01
feat(sdk): integration test for selective disclosure.

vkubiv:

### DIFF
--- a/test/integration/openid4ci_test.go
+++ b/test/integration/openid4ci_test.go
@@ -7,7 +7,10 @@ SPDX-License-Identifier: Apache-2.0
 package integration
 
 import (
+	"encoding/base64"
+	"encoding/json"
 	"fmt"
+	"strings"
 	"testing"
 
 	goapi "github.com/trustbloc/wallet-sdk/pkg/api"
@@ -34,12 +37,14 @@ import (
 
 func TestOpenID4CIFullFlow(t *testing.T) {
 	type test struct {
-		issuerProfileID string
-		issuerDIDMethod string
-		walletDIDMethod string
+		issuerProfileID       string
+		issuerDIDMethod       string
+		walletDIDMethod       string
+		isSelectiveDisclosure bool
 	}
 
 	tests := []test{
+		{issuerProfileID: "bank_issuer_jwtsd", issuerDIDMethod: "orb", walletDIDMethod: "ion", isSelectiveDisclosure: true},
 		{issuerProfileID: "bank_issuer", issuerDIDMethod: "orb", walletDIDMethod: "ion"},
 		{issuerProfileID: "did_ion_issuer", issuerDIDMethod: "ion", walletDIDMethod: "key"},
 	}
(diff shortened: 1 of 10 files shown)


test/integration/openid4ci_test.go
```diff
@@ -109,11 +114,36 @@ func TestOpenID4CIFullFlow(t *testing.T) {
 		require.NoError(t, err)
 		require.Contains(t, vc.VC.Issuer.ID, tc.issuerDIDMethod)
 
-		subID, err := verifiable.SubjectID(vc.VC.Subject)
-		require.Contains(t, subID, didID)
+		if !tc.isSD {
+			subID, err := verifiable.SubjectID(vc.VC.Subject)
+			require.NoError(t, err)
+			require.Contains(t, subID, didID)
+			checkActivityLogAfterOpenID4CIFlow(t, activityLogger, tc.issuerProfileID, subID)
```
rolsonquadras, line 121: Can we put this validation outside the if-else condition ?

sudeshrshetty (approved):

test/integration/openid4ci_test.go
vkubiv: it will not work, because we use selective disclosure there will not be an `id` inside credentialSubject. So it will not appear in the logs.

rolsonquadras: Ok. Do we need a follow up PR to fix the activity log for SD ?

rolsonquadras (approved):

test/integration/openid4ci_test.go
```diff
@@ -109,11 +114,36 @@ func TestOpenID4CIFullFlow(t *testing.T) {
...
+		} else {
+			subID := getDisclosureHelper(t, serializedVC, "id")
+			require.Contains(t, subID, didID)
```
DRK3, line 124: You can remove the else and move the body of the else outside

test/integration/openid4ci_test.go
```diff
@@ -37,9 +40,11 @@ func TestOpenID4CIFullFlow(t *testing.T) {
 		issuerProfileID string
 		issuerDIDMethod string
 		walletDIDMethod string
+		isSD            bool
```
DRK3, line 43: Consider renaming this to `isSelectiveDisclosure` or at least add a comment explaining what the `SD` stands for - I don't think this is an obvious acronym for future readers.

test/integration/openid4ci_test.go
vkubiv: getDisclosureHelper will fail for regular VC

DRK3 (approved):
Minor comments, please check

Moopli (approved):

test/integration/openid4ci_test.go
DRK3: My bad - when I wrote my comment, I thought your `if` block ended in a return, but I see now that it doesn't. What I was thinking of was this:
```
if !tc.isSelectiveDisclosure {
	subID, err := verifiable.SubjectID(vc.VC.Subject)
	require.NoError(t, err)
	require.Contains(t, subID, didID)
	checkActivityLogAfterOpenID4CIFlow(t, activityLogger, tc.issuerProfileID, subID)
	return
}

subID := getDisclosure(t, serializedVC, "id")
			require.Contains(t, subID, didID)
```

PR #223, opened 2023-02-01
fix(app): Implement sharedpreference storage for android to inject to localkms

talwinder50:

### DIFF
--- a/demo/app/android/app/src/main/kotlin/walletsdk/kmsStorage/kmsStore.kt
+++ b/demo/app/android/app/src/main/kotlin/walletsdk/kmsStorage/kmsStore.kt
@@ -1,0 +1,38 @@
+package walletsdk.kmsStorage
+import android.content.Context
+import android.os.Build
+import androidx.annotation.RequiresApi
+import dev.trustbloc.wallet.sdk.localkms.Result
+import dev.trustbloc.wallet.sdk.localkms.Store
+import java.util.*
+
+
+class KmsStore(context: Context) : Store {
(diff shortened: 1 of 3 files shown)


demo/app/android/app/src/main/kotlin/walletsdk/kmsStorage/kmsStore.kt
```diff
@@ -0,0 +1,38 @@
...
+        return localResult
+    }
+
+    @RequiresApi(Build.VERSION_CODES.O)
+    override fun put(keySetID: String?, key: ByteArray?) {
+        try{
+            val editor = preferences.edit()
+            editor.putString(keySetID, Base64.getEncoder().encodeToString(key))
+            editor.apply()
+        } catch(e: java.lang.Exception) {
```
DRK3, line 33: You should allow the exception to be passed on to the calling code instead of catching it here. If you catch it here, the calling code will think the `Put` call succeeded, preventing the error handling from working as expected.

DRK3 (approved):
Left one comment

vkubiv (approved):

rolsonquadras (approved):

demo/app/android/app/src/main/kotlin/walletsdk/kmsStorage/kmsStore.kt
rolsonquadras: @talwinder50 to raise a follow up PR.

PR #229, opened 2023-02-03
test(sdk): added driver's license issuer and test flow

birtony:
Signed-off-by: Anton Biriukov <anton.biriukov@avast.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ copy-android-bindings:
 
 .PHONY: copy-ios-bindings
 copy-ios-bindings:
-	@cp -R cmd/wallet-sdk-gomobile/bindings/ios/walletsdk.xcframework demo/app/ios/Runner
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -48,7 +48,7 @@ copy-android-bindings:
 
 .PHONY: copy-ios-bindings
 copy-ios-bindings:
-	@cp -R cmd/wallet-sdk-gomobile/bindings/ios/walletsdk.xcframework demo/app/ios/Runner
+	@rm -rf demo/app/ios/Runner/walletsdk.xcframework && cp -R cmd/wallet-sdk-gomobile/bindings/ios/walletsdk.xcframework demo/app/ios/Runner
```
DRK3, line 51: Reason for this change?

birtony: When previously generate locally, running this target the second time would throw an error since it can't override and existing folder. This might have to do with the local default permissions, but  I think cleaning the old binding doesn't hurt.

vkubiv (approved):

PR #242, opened 2023-02-08
refactor(sdk,app): unify jwt signing.

vkubiv:

### DIFF
--- a/cmd/wallet-sdk-gomobile/openid4ci/interaction.go
+++ b/cmd/wallet-sdk-gomobile/openid4ci/interaction.go
@@ -8,13 +8,8 @@ SPDX-License-Identifier: Apache-2.0
 package openid4ci
 
 import (
-	"encoding/json"
-	"fmt"
-
-	"github.com/btcsuite/btcutil/base58"
-	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
-	"github.com/hyperledger/aries-framework-go/pkg/doc/util/didsignjwt"
 	goapi "github.com/trustbloc/wallet-sdk/pkg/api"
+	"github.com/trustbloc/wallet-sdk/pkg/common"
 
 	"github.com/trustbloc/wallet-sdk/cmd/wallet-sdk-gomobile/api"
 	"github.com/trustbloc/wallet-sdk/cmd/wallet-sdk-gomobile/walleterror"
@@ -26,6 +21,7 @@ import (
 // object are used to help guide the calling code through the OpenID4CI flow.
 type Interaction struct {
 	goAPIInteraction *openid4cigoapi.Interaction
+	crypto           api.Crypto
 }
 
 // AuthorizeResult is the object returned from the Client.Authorize method.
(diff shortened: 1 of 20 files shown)


demo/app/android/app/src/main/kotlin/walletsdk/flutter/app/MainActivity.kt
```diff
@@ -195,7 +189,10 @@ class MainActivity : FlutterActivity() {
         val openID4CI = this.openID4CI
             ?: throw java.lang.Exception("openID4CI not initiated. Call authorize before this.")
 
-        return openID4CI.requestCredential(otp)
+        val didVerificationMethod = this.didVerificationMethod
+                ?: throw java.lang.Exception("DID should be created first")
+
+        return openID4CI.requestCredential(otp, didVerificationMethod)
```
rolsonquadras, line 195: API surface is changed with this PR. Can we update the documentation as well ?

rolsonquadras: Also, can we add Breaking changes tag to the commit message ?

vkubiv: yes, sure

Moopli (approved):

cmd/wallet-sdk-gomobile/openid4ci/interaction.go
```diff
@@ -50,23 +46,21 @@ func NewCredentialRequestOpts(userPIN string) *CredentialRequestOpts {
 // ActivityLogger is optional, but if provided then activities will be logged there.
 // If not provided, then no activities will be logged.
 type ClientConfig struct {
-	UserDID        string
 	ClientID       string
-	SignerCreator  api.DIDJWTSignerCreator
+	Crypto         api.Crypto
```
DRK3, line 50: The `api.Crypto` interface contains both `Sign` and `Verify` methods, but here it looks like we're only using the `Sign` method.  `api.Crypto` Should be broken down into `Signer` and `Verifier` interfaces so we don't require callers to implement unused methods.

vkubiv: we need to do this in separate PR. and discuss this with @sudeshrshetty . Because it is against the initial design.

demo/app/android/app/src/main/kotlin/walletsdk/flutter/app/MainActivity.kt
vkubiv: done

rolsonquadras (approved):

DRK3 (approved):

cmd/wallet-sdk-gomobile/openid4ci/interaction.go
sudeshrshetty: I'm ok, we can split it.

PR #246, opened 2023-02-09
draft feat(app): Persist the activity logger in the storage

talwinder50:


talwinder50:
Need to revisit this logic later

PR #253, opened 2023-02-13
fix(app): Fix the image not rendering in sdJWT flow

talwinder50:


DRK3 (approved):

talwinder50:
addressed in the follow up pr already

PR #265, opened 2023-02-21
feat(sdk): Update OpenID4CI to support 1_0-11

DRK3:
* Updated OpenID4CI code to support the latest 1_0-11 version.
* Removed the optional ResolveDisplay convenience method on the OpenID4CI interaction object as it doesn't seem to be useful - the standalone ResolveDisplay function still exists and can be used instead.


rolsonquadras (approved):

cmd/wallet-sdk-gomobile/docs/usage.md
```diff
@@ -347,7 +344,6 @@ interaction.authorize() // Returned object doesn't matter with current implement
 val userPIN = "1234"
 val requestCredentialOpts = CredentialRequestOpts(userPIN)
 val credentials = interaction.requestCredential(requestCredentialOpts, doc.assertionMethod()) // Should probably store these somewhere
```
Moopli, line 346: While we're at it, I noticed that this line refers to `doc` while line 337 above sets `didDocResolution`.

DRK3: Good catch - fixed.

pkg/credentialschema/testdata/issuer_metadata.json
```diff
@@ -1,21 +1,26 @@
...
-    "university_degree":{
-      "claims":{
+  "display":[
+    {
+      "locale":"en-US",
+      "name":"Example University"
+    },
+    {
+      "name":"サンプル大学",
+      "locale":"jp-JA"
```
birtony, line 12: Shouldn't `locale` come as a second key?

talwinder50 (approved):

Moopli (approved):

pkg/credentialschema/testdata/issuer_metadata.json
DRK3: The order doesn't matter, but it would look nicer if it was consistent. Next time I modify that file I'll update it

PR #279, opened 2023-02-26
feat(sdk)!: Metrics logging, ResolveDisplay refactor

DRK3:
BREAKING CHANGE: The openid4ci.ResolveDisplay function has been moved to its own package to better reflect how it can be called independent of the OpenID4CI flow and to improve the generated names in gomobile. The parameters for the ResolveDisplay function have been moved inside of a new struct type called ResolveDisplayOpts to make it easier to add new parameters in the future without breaking the API.


sudeshrshetty (approved):

vkubiv (approved):

rolsonquadras (approved):

cmd/wallet-sdk-gomobile/api/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+
+// A MetricsEvent represents a single event that occurred that had performance metrics taken.
+type MetricsEvent struct {
+	// GoAPIMetricsEvent will not be accessible directly in the bindings (will be "skipped").
+	// This is OK - this field is not intended to be set directly by the user of the SDK.
+	GoAPIMetricsEvent *goapi.MetricsEvent
+}
+
+// Event returns the event name.
+func (m *MetricsEvent) Event() string {
```
birtony, line 19: Just curious if you think it'd be cleaner to have function names start with a verb to better represent what they're doing? `getEvent()`, etc

DRK3: My thinking was that for a function that's a simple "getter" type of operation that doesn't require DB lookups or other more complex operations, that a simple noun should suffice. If this call required a database lookup or something like that, I think `GetEvent()` would make more sense. Regardless of that, currently in the SDK all the other methods that do similar things (simple field returns) don't use `get`, so I think it's best to keep it consistent. Thanks for providing the suggestion, though - let me know what you think of my reasonin

birtony: Fair enough, thanks @DRK3! 👍

rolsonquadras:
Merging this as Android build fails in main branch as well - fix to android test in a follow up PR.

PR #281, opened 2023-02-27
feat(sdk): Mobile VC types, issuance date, and expiration date methods

DRK3:


vkubiv (approved):

cmd/wallet-sdk-gomobile/api/credentials.go
```diff
@@ -39,6 +41,37 @@ func (v *VerifiableCredential) IssuerID() string {
...
+func (v *VerifiableCredential) IssuanceDate() (int64, error) {
+	if v.VC.Issued == nil {
+		return -1, errors.New("issuance date missing (invalid VC)")
+	}
+
+	return v.VC.Issued.Unix(), nil
+}
+
+// HasExpirationDate returns whether this VC has an expiration date.
+func (v *VerifiableCredential) HasExpirationDate() bool {
```
rolsonquadras, line 60: Do we need this function as I see ExpirationDate() returns -1 if its nil ?

rolsonquadras (approved):

cmd/wallet-sdk-gomobile/api/credentials.go
```diff
@@ -39,6 +41,37 @@ func (v *VerifiableCredential) IssuerID() string {
 	return v.VC.Issuer.ID
 }
 
+// Types returns the types of this VC. At a minimum, one of the types will be "VerifiableCredential".
+// There may be additional more specific credential types as well.
+func (v *VerifiableCredential) Types() *StringArray {
+	return &StringArray{strings: v.VC.Types}
```
rolsonquadras, line 47: @sudeshrshetty do we need to pick Types from credentialSubject as well ?

sudeshrshetty: Either we can add a function to get `Claim() returning credential subject map` & `ClaimType() which returns credential subject type if found`

sudeshrshetty: can be a new PR

cmd/wallet-sdk-gomobile/api/credentials.go
DRK3: I considered that, but ultimately decided that we should have the `HasExpirationDate` function because:

1. Unix timestamps can be negative - while it probably doesn't make much sense for any realistic credential to be issued with a date that would result in a negative unix time (meaning earlier than 1 January 1970), it is a technical possibility
2. To avoid requiring the caller to understand that -1 implicitly means "no timestamp" - having the explicit method + error return makes it more clear in my opinion.

PR #287, opened 2023-02-28
refactor: flutter plugin ios

vkubiv:


demo/app/ios/Runner/OpenID4CI.swift
```diff
@@ -0,0 +1,45 @@
+//
+//  OpenID4CI.swift
+//  Runner
+//
+//  Created by Volodymyr Kubiv on 28.02.2023.
+//
+
```
birtony, line 7: Are we not including those copyright headers in swift files?

rolsonquadras: Looks like we have custom license headers for iOS/Android files. Let's have a follow up PR to update license headers.

rolsonquadras (approved):

talwinder50 (approved):

birtony (approved):

PR #309, opened 2023-03-07
feat(sdk): Support for masking, raw ID, and pattern data in display

DRK3:
Note that the masking implementation only supports regex that specifies characters at the beginning of a string to mask.

### DIFF
--- a/cmd/wallet-sdk-gomobile/display/displaydata.go
+++ b/cmd/wallet-sdk-gomobile/display/displaydata.go
@@ -10,6 +10,7 @@ package display
 import (
 	"encoding/json"
 	"errors"
+	"strings"
 
 	goapicredentialschema "github.com/trustbloc/wallet-sdk/pkg/credentialschema"
 )
@@ -193,6 +194,12 @@ type Claim struct {
 	claim *goapicredentialschema.ResolvedClaim
 }
 
+// RawID returns the raw field name (key) from the VC associated with this claim.
+// It's not localized or formatted for display.
+func (c *Claim) RawID() string {
+	return c.claim.RawID
+}
+
 // Label returns the display label for this claim.
 // For example, if the UI were to display "Given Name: Alice", then the Label would be "Given Name".
 func (c *Claim) Label() string {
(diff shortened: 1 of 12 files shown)


birtony (approved):

vkubiv (approved):

rolsonquadras (approved):

cmd/wallet-sdk-gomobile/display/displaydata.go
```diff
@@ -211,6 +218,30 @@ func (c *Claim) Value() string {
 	return c.claim.Value
 }
 
+// RawValue returns the raw display value for this claim without any formatting.
+// For example, if this claim is masked, this method will return the unmasked version.
+// If no special formatting was applied to the display value, then this method will be equivalent to calling Value.
+func (c *Claim) RawValue() string {
+	if c.claim.RawValue == "" {
```
rolsonquadras, line 225: Can this ever be empty and display value not empty ? We are deriving display value from raw value.

sudeshrshetty: yes, we don't need this check

DRK3: With the way I've implemented it, there are currently exactly three scenarios for a claim object:

1. Claim is not masked: `Value()` and `RawValue()` do the exact same thing - return the display value.
2. Claim is masked: `Value()` returns the masked display value, `RawValue()` returns the unmasked display value.
3. Bad/incompatible issuer metadata received: `Value()` and `RawValue()` do the exact same thing, and neither one is masked (this is the "default display" route).

Nothing else should be possible. Note: we may want to consider changing how scenario 3 is handled.

DRK3: For most claims, there's no difference between the value and raw value. With this in mind, I didn't want to pollute the serialized version of the display data (`c.claim`) with duplicate information. So, in order to save space, I only set `claim.RawID` when necessary. See https://github.com/trustbloc/wallet-sdk/pull/309/files#diff-7a58209a9768b5b21b69afaf26b3a88c52dd9a4376ce53e049d83d23afbd410fR188

talwinder50 (approved):

cmd/wallet-sdk-gomobile/display/displaydata.go
DRK3: To make the code clearer, I changed this so that internally (in the "Go SDK" layer), the `RawValue` field is the one that's always set, and the `Value` field is what's conditionally set when the value has formatting (e.g. masking).

Before, what happened was that for data that didn't have any formatting (always "raw"), `Value` would be set, and for formatted data, `Value` became the formatted version while `RawValue` would be the unformatted one.

Now in the gomobile layer, the `Value` method is the one that has the if condition.

The only difference is in terms of how the internal data model looked - SDK usage doesn't change, but now the code is more intuitive.

sudeshrshetty (approved):

PR #310, opened 2023-03-07
fix(app): fixed an issue with adv ports not being reversed in CI

birtony:
Signed-off-by: Anton Biriukov <anton.biriukov@avast.com>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -170,7 +170,10 @@ jobs:
           force-avd-creation: false
           emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
           disable-animations: false
-          script: echo "Generated AVD snapshot for caching."
+          script: |
+            adb reverse tcp:8075 tcp:8075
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -170,7 +170,10 @@ jobs:
           force-avd-creation: false
           emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
           disable-animations: false
-          script: echo "Generated AVD snapshot for caching."
+          script: |
+            adb reverse tcp:8075 tcp:8075
```
rolsonquadras, line 174: @birtony as discussed, looks like this change is trying to fix intermittent error https://github.com/trustbloc/wallet-sdk/actions/runs/4345259748/jobs/7595098735#step:14:152. But the actual error is coming from golang cli tool than the emulator. Similar error shows up in iOS tests as well -https://github.com/trustbloc/wallet-sdk/actions/runs/4326039936/jobs/7552969166#step:10:8. The issue is probably with docker containers not starting.

PR #325, opened 2023-03-13
fix(app): Refactor flutter logic to the sdk display api

talwinder50:


demo/app/lib/widgets/credential_verified_information_view.dart
```diff
@@ -17,20 +16,16 @@ class CredentialVerifiedInformation extends StatelessWidget {
...
-    var credentialClaimsData = data['credential_displays'][0]['claims'] as List;
+    var credentialDisplayData  = await WalletSDKPlugin.resolveCredDisplayRendering(credentialData.credentialDisplayData);
+    var credentialDisplayEncodeData = json.encode(credentialDisplayData);
+    List<dynamic> responseJson = json.decode(credentialDisplayEncodeData);
+    var credentialClaimsData = responseJson.first['claims'] as List;
     list = credentialClaimsData.map<CredentialPreviewData>((json) => CredentialPreviewData.fromJson(json)).toList();
-    if (hasOrder == true){
-      log("hasOrder $hasOrder");
-      list.sort((a, b) {
+    // TODO order function in the sdk is not working properly
```
rolsonquadras, line 24: Can you provide more info? Are the values not returned from the function?

talwinder50: @rolsonquadras  - `(BOOL)order:(long* _Nullable)ret0_ error:(NSError* _Nullable* _Nullable)error;` This is the function defination. As per golang implementation it should return int or long but rather i keep getting error () is not type of int. When i add this in the swift app crashes complaining there is no () type. Its empty parathenthis which i have not seen before.

rolsonquadras: ^ @DRK3

talwinder50: https://github.com/trustbloc/wallet-sdk/issues/326 issue is created

demo/app/ios/Runner/flutterPlugin.swift
```diff
@@ -339,6 +342,51 @@ public class SwiftWalletSDKPlugin: NSObject, FlutterPlugin {
...
+        let displayData = DisplayParseData(resolvedCredentialDisplayData, nil)
+        let issuerDisplayData = displayData?.issuerDisplay()
+        
+        var resolvedCredDisplayList : [Any] = []
+        var claimList:[Any] = []
+        
+        for i in 0...((displayData!.credentialDisplaysLength())-1){
+            let credentialDisplay = displayData!.credentialDisplay(at: i)
+            for i in 0...(credentialDisplay!.claimsLength())-1{
+                let claim = credentialDisplay?.claim(at: i)
```
DRK3, line 363: Consider simply force unwrapping `credentialDisplay`, as it will be never be nil as long as the loop logic is correct (in Wallet-SDK, I only return `nil` if you attempt to get an object at an out of bounds index)

demo/app/ios/Runner/flutterPlugin.swift
```diff
@@ -339,6 +342,51 @@ public class SwiftWalletSDKPlugin: NSObject, FlutterPlugin {
...
+            resolveDisplayResp["overviewName"] = overview?.name()
+            resolveDisplayResp["logo"] = logo?.url()
+            resolveDisplayResp["textColor"] = overview?.textColor()
+            resolveDisplayResp["backgroundColor"] = overview?.backgroundColor()
+            resolveDisplayResp["issuerName"] = issuerDisplayData!.name()
+            
+            resolvedCredDisplayList.append(resolveDisplayResp)
+        }
+    
+        print("resolvedCredDisplay ->", resolvedCredDisplayList)
```
DRK3, line 386: Intended to leave this in?

demo/app/lib/demo_method_channel.dart
```diff
@@ -141,6 +141,12 @@ class MethodChannelWallet extends WalletPlatform {
       return hasOrder;
   }
 
+  Future<List<Object?>> resolveCredDisplayRendering(String resolvedCredentialDisplayData) async {
+    log("inside rendering");
```
DRK3, line 145: Intended to leave these log statements in?

rolsonquadras (approved):

demo/app/lib/widgets/credential_verified_information_view.dart
DRK3: Have to pass in a `UnsafeMutablePointer<Int>` as the first argument. It is weird, but it's likely for some sort of compatibility reason during the Go<->Obj C translation.

DRK3 (commented):
Left some comments

vkubiv (approved):

demo/app/ios/Runner/flutterPlugin.swift
```diff
@@ -49,6 +49,9 @@ public class SwiftWalletSDKPlugin: NSObject, FlutterPlugin {
         case "resolveCredentialDisplay":
             resolveCredentialDisplay(arguments: arguments!,  result: result)
             
+        case "credentialDisplayRendering":
```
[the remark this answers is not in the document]
talwinder50: Its the naming which i will rename, we want to persist the serliazed credential display data in the app. So earlier i was using that json to render to the ui but now i pass that to this second func credentialDisplayRendering. I will rename it properly in the follow up pr #326

demo/app/ios/Runner/flutterPlugin.swift
talwinder50: removed

demo/app/ios/Runner/flutterPlugin.swift
```diff
@@ -339,6 +342,51 @@ public class SwiftWalletSDKPlugin: NSObject, FlutterPlugin {
...
+                let claim = credentialDisplay?.claim(at: i)
+                var claims : [String: Any] = [:]
+                claims["rawValue"] = claim?.rawValue()
+                claims["value"] = claim?.value()
+                claims["valueType"] = claim?.valueType()
+                claims["label"] = claim?.label()
+                claimList.append(claims)
+            }
+     
+            let overview = credentialDisplay?.overview()
```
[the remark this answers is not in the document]
talwinder50: updated

demo/app/ios/Runner/flutterPlugin.swift
talwinder50: updated

DRK3 (approved):

demo/app/ios/Runner/flutterPlugin.swift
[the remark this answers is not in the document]
DRK3: My recommendation - change `resolveCredentialDisplay` to `serializeDisplayData` and `credentialDisplayRendering` to `resolveCredentialDisplay` or `renderCredentialDisplay` or something like that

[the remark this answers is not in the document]
talwinder50: yea doing that in #326

PR #333, opened 2023-03-15
fix(app): Fix ordering and loading issue

talwinder50:


demo/app/ios/Runner/flutterPlugin.swift
```diff
@@ -367,6 +366,15 @@ public class SwiftWalletSDKPlugin: NSObject, FlutterPlugin {
                      claims["value"] = claim.value()
                      claims["rawValue"] = claim.rawValue()
                 }
+                var order: Int = -1
+                if claim.hasOrder() == true {
+                    do {
+                        try claim.order(&order)
```
DRK3, line 372: Question: if you do

```
claim.order(&order, nil)
```
does the compiler allow you to get rid of the try/catch? The reason I ask is because if `hasOrder` returns true then it's safe to skip checking the error for `order`.

demo/app/ios/Runner/flutterPlugin.swift
```diff
@@ -367,6 +366,15 @@ public class SwiftWalletSDKPlugin: NSObject, FlutterPlugin {
                      claims["value"] = claim.value()
                      claims["rawValue"] = claim.rawValue()
                 }
+                var order: Int = -1
+                if claim.hasOrder() == true {
```
DRK3, line 370: Can be simplified to `if claim.hasOrder() {`

DRK3 (approved):
Minor comments

demo/app/ios/Runner/flutterPlugin.swift
talwinder50: No it gives compile time error

DRK3: (see my documentation for more info: https://github.com/trustbloc/wallet-sdk/blob/main/cmd/wallet-sdk-gomobile/docs/usage.md#claim)

rolsonquadras (approved):

PR #350, opened 2023-03-22
fix(app): Add the version and show diddocument in the dev mode

talwinder50:
IMG_0537

### DIFF
--- a/demo/app/ios/Runner/flutterPlugin.swift
+++ b/demo/app/ios/Runner/flutterPlugin.swift
@@ -48,20 +48,22 @@ public class SwiftWalletSDKPlugin: NSObject, FlutterPlugin {
 
         case "serializeDisplayData":
             serializeDisplayData(arguments: arguments!,  result: result)
-            
+
         case "resolveCredentialDisplay":
             resolveCredentialDisplay(arguments: arguments!, result: result)
 
-            
+        case "getVersionDetails":
+           getVersionDetails(result:result)
+
         case "getCredID":
             getCredID(arguments: arguments!,  result: result)
-            
+
         case "parseActivities":
             parseActivities(arguments: arguments!,  result: result)
-            
+
         case "initSDK":
             initSDK(result:result)
-            
+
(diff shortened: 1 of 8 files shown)


demo/app/ios/Runner/flutterPlugin.swift
```diff
@@ -67,6 +67,9 @@ public class SwiftWalletSDKPlugin: NSObject, FlutterPlugin {
 
         case "activityLogger":
             storeActivityLogger(result:result)
+        
+        case "getVersionDetail":
```
vkubiv, line 71: maybe rename it to getWalletSDKVersion or something like this, to not confuse it with the app version

talwinder50: yea make sense will do that

demo/app/ios/Runner/flutterPlugin.swift
```diff
@@ -89,7 +92,16 @@ public class SwiftWalletSDKPlugin: NSObject, FlutterPlugin {
...
-    
+     
+    /**
+              This method gets the version detail if we build sdk using the env variable
+              For Example: NEW_VERSION=testVer GIT_REV=testRev BUILD_TIME=testTime make generate-ios-bindings copy-ios-bindings
+     */
+    public func getWalletSDKVersion(result: @escaping FlutterResult) {
+        let version = VersionGetVersion()
+        print("getting version detail ->", version)
+        result(version)
```
rolsonquadras, line 103: Can we show Gitversion as well as Build time as well ?

https://github.com/trustbloc/wallet-sdk/blob/main/demo/app/ios/Runner/Tests/IntegrationTest.swift#L25-L26

talwinder50: Yes can add that too. Will add n update

rolsonquadras (approved):

DRK3 (approved):
Looks good, would suggest renaming `getVersionDetails` as @birtony suggested before merging

PR #354, opened 2023-03-23
feat(sdk): Injectable LD document loader, better test stability

### LINKED ISSUE #160: Support custom document loaders
Based on TODO https://github.com/trustbloc/wallet-sdk/blob/ae3270b775cce256cb3762644933172c71c79ad7/cmd/wallet-sdk-gomobile/credential/memstorage.go#L30

DRK3:
* LD document loaders can now be injected wherever LD document loaders are used.
* Updated default document loader across the SDK to be consistent: now, whenever a document loader is not specified (or is nil), a network-based loader will be used by default.
* Updated docs to described this new default LD loader behaviour.
* Updated unit tests and BDD tests to ensure they never use a network-based loader in order to improve test reliability and performance. Now, tests either have document loading disabled or use pre-loaded contexts. Note that Flutter tests will still use network-based loading for the time being.

closes #160

### DIFF
--- a/pkg/openid4ci/clientconfig.go
+++ b/pkg/openid4ci/clientconfig.go
@@ -10,11 +10,14 @@ import (
 	"errors"
 	"net/http"
 
-	"github.com/trustbloc/wallet-sdk/pkg/api"
-	"github.com/trustbloc/wallet-sdk/pkg/walleterror"
-
 	diddoc "github.com/hyperledger/aries-framework-go/pkg/doc/did"
 	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api/vdr"
+	"github.com/piprate/json-gold/ld"
+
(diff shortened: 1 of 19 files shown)


pkg/openid4ci/clientconfig.go
```diff
@@ -66,3 +70,21 @@ func validateClientConfig(config *ClientConfig) error {
 
 	return nil
 }
+
+func setDefaultsAsNeeded(config *ClientConfig) {
```
rolsonquadras, line 74: I feel setDefault is self explanatory. ie, without the suffix "AsNeeded".

rolsonquadras (approved):

pkg/openid4ci/clientconfig.go
DRK3: Done - changed to `setDefaults`

vkubiv (approved):

PR #358, opened 2023-03-24
feat(sdk)!: Improved gomobile API parameter design

DRK3:
* Reworked the gomobile API parameters to make it easier to add new optional parameters in the future without causing breaking changes, while also being more consistent + easier to understand across Wallet-SDK.
* Removed the network-based LD document loader implementation from the gomobile SDK as it's completely redundant now. The default for all APIs is to use a network-based document loader whenever one isn't specified.
* Removed the DIDCreator interface since it's not used anywhere. The options for the Create DID method have been moved from the api package to the did package.


cmd/wallet-sdk-gomobile/docs/usage.md
```diff
@@ -234,7 +230,7 @@ import dev.trustbloc.wallet.sdk.localkms.MemKMSStore
 val memKMSStore = MemKMSStore.MemKMSStore()
 val kms = Localkms.newKMS(memKMSStore)
 val didCreator = Creator(kms as KeyWriter)
-val didDocResolution = didCreator.create("key", CreateDIDOpts()) // Create a did:key doc
+val didDocResolution = didCreator.create("key", null) // Create a did:key doc
```
rolsonquadras, line 233: I understand we internally create default opts in case of null, but can we have examples with Opts passed to the API? Same comments for other examples.

rolsonquadras (commented):
Can add a section to give an overview of the API design structure 
(required args vs optional args) ?

Moopli (approved):

DRK3:
> Can add a section to give an overview of the API design structure (required args vs optional args) ?

Done!

cmd/wallet-sdk-gomobile/docs/usage.md
DRK3: Sure - but I think I should do it in a follow-up PR, since it's a little bit out of scope and this PR has already gotten pretty huge.

DRK3:
I've updated the Swift examples to pass in `NSError` objects into any functions that require it as an input parameter.

vkubiv (approved):

demo/app/android/app/src/androidTest/kotlin/integration_test.kt
```diff
@@ -3,20 +3,15 @@ import androidx.test.filters.SmallTest
 import androidx.test.platform.app.InstrumentationRegistry
 import com.google.common.truth.Truth.assertThat
 import dev.trustbloc.wallet.BuildConfig
-import dev.trustbloc.wallet.sdk.api.CreateDIDOpts
 import dev.trustbloc.wallet.sdk.api.KeyWriter
 import dev.trustbloc.wallet.sdk.api.VerifiableCredentialsArray
-import dev.trustbloc.wallet.sdk.credential.CredentialsOpt
+import dev.trustbloc.wallet.sdk.credential.CredentialsArg
```
rolsonquadras, line 8: I see other options ending with plural (`zzzArgs`), and this one is singular (`CredentialsArg`). What's the reason for this?

rolsonquadras: Also, I see 2 suffices - zzzRequiredArgs and zzzOptionalArgs. This one match to that standards.

DRK3: I only put `required` in the name for a few places where there are many different required input arguments (OpenID4CI/OpenID4VP), but really any name could do. I think a better solution in the future would be to try to defer passing in those extra arguments until they're really needed, and that would make those big argument bundles less necessary.

rolsonquadras (approved):

PR #359, opened 2023-03-27
feat(sdk): add support of json-ld to vp flow.

vkubiv:


DRK3 (approved):
Since there are no changes to non-test code, I suppose this should be `test` instead of `feat` in the commit title

birtony (approved):

PR #360, opened 2023-03-28
fix(app): Add status api ui change

talwinder50:


demo/app/lib/widgets/credential_metadata_card.dart
```diff
@@ -101,7 +109,36 @@ class CredentialMetaDataCard extends StatefulWidget {
...
                           ),
                           textAlign: TextAlign.start,
-                        )
+                        ),
+                      trailing: verifiedStatus ? const Text.rich(
+                        TextSpan(
+                          children: [
+                            WidgetSpan(child: Icon(Icons.verified_user_outlined,color: Colors.lightGreen, size: 18,)),
+                           TextSpan(
+                           text: 'Verified',
```
rolsonquadras, line 118: The status check API says whether the credential is active or revoked. Can we change the text to active instead of verified?

vkubiv (approved):

demo/app/lib/widgets/credential_metadata_card.dart
talwinder50: updated

rolsonquadras (approved):

PR #362, opened 2023-03-29
fix(sdk): DID Well-Known config validate

rolsonquadras:
- Pull latest AFG changes
- Trim slash suffix from the url before passing to AFG function


birtony (approved):

birtony (commented):
Should we add the new copyright heading to this file since it's modified?

birtony (approved):

talwinder50 (approved):

PR #378, opened 2023-04-04
feat(sdk)!: Improve names in credential package in generated bindings

DRK3:


talwinder50 (approved):

vkubiv (approved):

demo/app/ios/Runner/OpenID4VP.swift
```diff
@@ -67,7 +67,7 @@ public class OpenID4VP {
             throw OpenID4VPError.runtimeError("OpenID4VP interaction not properly initialized, call processAuthorizationRequest first")
         }
         
-//         let  verifiablePresentation = try CredentialNewInquirer(documentLoader)!.query(vpQueryContent, contents: CredentialCredentialsOpt(selectedCredentials))
+//         let  verifiablePresentation = try CredentialNewInquirer(documentLoader)!.query(vpQueryContent, credentials: selectedCredentials)
```
birtony, line 70: I see this commented code was here before and is not a part of your changes, but just wondering why is it here?

birtony (approved):
I noticed some of the updated files are missing copyright headers, might need to add those as well

rolsonquadras (approved):

DRK3:
> I noticed some of the updated files are missing copyright headers, might need to add those as well

Yeah, I think most of the demo app files are missing them - I guess we should add them at some point CC: @rolsonquadras

PR #388, opened 2023-04-07
feat(sdk): verifier display data.

vkubiv:


pkg/openid4vp/openid4vp.go
```diff
@@ -133,10 +133,36 @@ func (o *Interaction) GetQuery() (*presexch.PresentationDefinition, error) {
...
+		return nil, walleterror.NewExecutionError(
+			module,
+			NotInitializedProperlyErrorCode,
+			NotInitializedProperlyError,
+			fmt.Errorf("call GetQuery first"))
+	}
+
+	return &VerifierDisplayData{
+		DID:     o.requestObject.ClientID,
+		Name:    o.requestObject.Registration.ClientName,
```
rolsonquadras, line 148: Is registration data mandatory in response data ? If no, will there be a nil pointer if registration data is missing here?

vkubiv: @rolsonquadras 
requestObject contains Registration as a value, so it can't be null. 
`	Registration requestObjectRegistration `json:"registration"``

If registration is missed in json, ClientName and other fields will be just empty strings.

rolsonquadras (approved):

talwinder50 (approved):

PR #400, opened 2023-04-20
feat(sdk): bbs+ support.

vkubiv:


Moopli (approved):

talwinder50 (approved):

pkg/credentialquery/credentialquery.go
```diff
@@ -77,8 +77,6 @@ func (c *Instance) GetSubmissionRequirements(
 	results, err := query.MatchSubmissionRequirement(
 		credentials,
 		c.documentLoader,
-		verifiable.WithDisabledProofCheck(),
```
DRK3, line 80: Reason for the removal of `verifiable.WithDisabledProofCheck()`? Do proof checks never happen in `MatchSubmissionRequirement`? If there is the potential for proof checking to happen here, then there should be an option to disable it still since it's an expensive operation and often times we don't need to proof check again if we know it was done already (e.g. VCs from the OpenID4CI flow)

vkubiv: MatchSubmissionRequirement no more directly parses VC. Now this function takes just 2 args( credentials, c.documentLoader), no more optional args.

PR #416, opened 2023-05-03
feat(app): daily app build with maven wallet-sdk.

vkubiv:


demo/app/android/app/build.gradle
```diff
@@ -22,6 +22,7 @@ if (flutterVersionName == null) {
 }
 
 def walletSdkPkgUsr = localProperties.getProperty('wallet-sdk-pkg.usr') ?: System.getenv('WALLET_SDK_USR')
+def walletSdkPkgVer = localProperties.getProperty('wallet-sdk-pkg.ver') ?: System.getenv('WALLET_SDK_VER')
```
DRK3, line 25: Is this supposed to be in here twice? (line 24 and 25)

DRK3: Nevermind - just realized it's `USR` and `VER`

DRK3 (approved):

.github/workflows/daily.yml
```diff
@@ -0,0 +1,94 @@
+#
+# Copyright Avast Software. All Rights Reserved.
```
birtony, line 2: I think this should be
```suggestion
# Copyright Gen Digital Inc. All Rights Reserved.
```

.github/workflows/daily.yml
```diff
@@ -0,0 +1,94 @@
+#
```
birtony, line 1: `daily.yml` sounds a bit confusing to me

.github/workflows/daily.yml
```diff
@@ -0,0 +1,94 @@
...
+# Copyright Avast Software. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+name: "wallet-sdk ci"
```
birtony, line 7: Maybe this could be `Wallet-SDK Daily Build` or something like that?

birtony (commented):
Added a couple of minor comments on naming conventions, which shouldn't be a blocker. Although, we should probably have the copyright text updated before this gets merged. Looks good otherwise!

.github/workflows/daily.yml
vkubiv: I didn't come up with better name

birtony (approved):

.github/workflows/daily.yml
DRK3: @birtony Do you have any suggestions? Were you thinking of something more descriptive, perhaps `dailybuild.yml`?

birtony: Something like that, @DRK3

PR #440, opened 2023-05-30
feat(sdk): OAuth2 Dynamic Client Registration API

DRK3:
Conforms to https://datatracker.ietf.org/doc/html/rfc7591.

Also added new methods on the Interaction object that a caller can use to determine whether an issuer support dynamic client registration (per https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata).

### DIFF
--- a/cmd/wallet-sdk-gomobile/oauth2/clientmetadata.go
+++ b/cmd/wallet-sdk-gomobile/oauth2/clientmetadata.go
@@ -1,0 +1,206 @@
+/*
+Copyright Gen Digital Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package oauth2
+
+import (
+	"github.com/hyperledger/aries-framework-go/component/kmscrypto/doc/jose/jwk"
+	"github.com/trustbloc/wallet-sdk/cmd/wallet-sdk-gomobile/api"
+	goapi "github.com/trustbloc/wallet-sdk/pkg/api"
+	goapioauth2 "github.com/trustbloc/wallet-sdk/pkg/oauth2"
+)
+
+// ClientMetadata represents a set of client metadata values.
+type ClientMetadata struct {
+	goAPIClientMetadata *goapioauth2.ClientMetadata
+}
+
+// NewClientMetadata creates a new ClientMetadata object.
+func NewClientMetadata() *ClientMetadata {
+	return &ClientMetadata{goAPIClientMetadata: &goapioauth2.ClientMetadata{}}
+}
+
+// RedirectURIs returns the redirect URIs.
+func (c *ClientMetadata) RedirectURIs() *api.StringArray {
+	return &api.StringArray{Strings: c.goAPIClientMetadata.RedirectURIs}
+}
+
+// SetRedirectURIs sets the redirect URIs.
+func (c *ClientMetadata) SetRedirectURIs(redirectURIs *api.StringArray) {
+	if redirectURIs == nil {
+		redirectURIs = api.NewStringArray()
+	}
+
+	c.goAPIClientMetadata.RedirectURIs = redirectURIs.Strings
+}
+
+// TokenEndpointAuthMethod returns the token endpoint authentication method.
+func (c *ClientMetadata) TokenEndpointAuthMethod() string {
+	return c.goAPIClientMetadata.TokenEndpointAuthMethod
+}
+
+// SetTokenEndpointAuthMethod sets the token endpoint authentication method.
+func (c *ClientMetadata) SetTokenEndpointAuthMethod(tokenEndpointAuthMethod string) {
+	c.goAPIClientMetadata.TokenEndpointAuthMethod = tokenEndpointAuthMethod
+}
+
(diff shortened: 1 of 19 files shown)


pkg/oadcrp/models.go
```diff
@@ -0,0 +1,33 @@
...
+
+// RegisterClientRequest represents a request to register an OAuth2 client.
+type RegisterClientRequest struct {
+	ClientName              string   `json:"client_name,omitempty"`
+	ClientURI               string   `json:"client_uri,omitempty"`
+	RedirectURIs            []string `json:"redirect_uris,omitempty"`
+	GrantTypes              []string `json:"grant_types,omitempty"`
+	ResponseTypes           []string `json:"response_types,omitempty"`
+	Scope                   string   `json:"scope,omitempty"`
+	TokenEndpointAuthMethod string   `json:"token_endpoint_auth_method,omitempty"`
```
sudeshrshetty, line 17: I see policy uri & jwks missing here. 
I understand it is hard to support each and every request param, so you can add custom parameter here (of type interface{}), will be useful during interop to provide additional client request param that may come up.

sudeshrshetty (commented):
I would suggest renaming package and also consider the fact that wallet SDK is going to support few additional schema of OAuth Registration going forward.

sudeshrshetty (commented):
I would suggest renaming package and also consider the fact that wallet SDK is going to support few additional schema of OAuth Registration going forward.

pkg/oadcrp/models.go
DRK3: I've added the missing params. It was easier to add in each one individually instead of trying to generalize it + add custom marshaller + support it via gomobile. However, if we do eventually need to add custom parameters that are outside of the core spec, then I can add that later

DRK3:
> I would suggest renaming package and also consider the fact that wallet SDK is going to support few additional schema of OAuth Registration going forward.

I've changed the package name to `oauth2`. I think this results in good names across Go, Java/Kotlin, and Objc/Swift.
in Kotlin, the import package is `oauth2`, and in Objc/Swift the methods + types are prefixed with OAuth2, so the `RegisterClient` function looks something like this: `Oauth2RegisterClient`

pkg/oauth2/models.go
```diff
@@ -0,0 +1,35 @@
...
+	ResponseTypes           []string `json:"response_types,omitempty"`
+	ClientName              string   `json:"client_name,omitempty"`
+	ClientURI               string   `json:"client_uri,omitempty"`
+	LogoURI                 string   `json:"logo_uri,omitempty"`
+	Scope                   string   `json:"scope,omitempty"`
+	Contacts                []string `json:"contacts,omitempty"`
+	TOSURI                  string   `json:"tos_uri,omitempty"`
+	PolicyURI               string   `json:"policy_uri,omitempty"`
+	JWKSURI                 string   `json:"jwks_uri,omitempty"`
+	JWKS                    string   `json:"jwks,omitempty"`
```
sudeshrshetty, line 23: jwks is an ojbect

cmd/wallet-sdk-gomobile/oauth2/clientmetadata.go
```diff
@@ -0,0 +1,206 @@
...
+	return c.goAPIClientMetadata.JWKSURI
+}
+
+// SetJWKSURI sets the JWK Set document URI.
+func (c *ClientMetadata) SetJWKSURI(jwksURI string) {
+	c.goAPIClientMetadata.JWKSURI = jwksURI
+}
+
+// JWKS returns the JWK Set document value. If the client metadata doesn't have one, then nil is returned instead.
+func (c *ClientMetadata) JWKS() *api.JSONWebKeySet {
```
vkubiv, line 163: Maybe better to rename it to JWKSet. So it is more clear that is a set of JWKs

vkubiv (approved):

cmd/wallet-sdk-gomobile/oauth2/clientmetadata.go
DRK3: Done!

DRK3:
I've updated the `RegisterClient` methods to take in the full registration endpoint instead of the authorization server's base URL. This will make it simpler for a user to simply pass in the `registration_endpoint` from OpenID Provider Metadata (see here for reference: https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata)

rolsonquadras (approved):

cmd/wallet-sdk-gomobile/oauth2/clientmetadata.go
vkubiv: we still have JWKS function, maybe better to rename it to JWKSet?

vkubiv (approved):

cmd/wallet-sdk-gomobile/oauth2/clientmetadata.go
DRK3: Oops, my mistake. I thought I had changed them. They should be changed now.

Moopli (approved):

PR #453, opened 2023-06-13
fix(sdk,app)!: performance of json-ld VCs

vkubiv:
Loading JSON-LD contexts takes a lot of time, now by default document loader with embedded contexts is used.

Breaking change: CredentialNewInquirer can now return an error as the last argument.

### Before

Swift: 
`let inquirer = CredentialNewInquirer(nil)`

### Now

Swift: 
```
var newInquirerError: NSError?
let inquirer = CredentialNewInquirer(nil, &newInquirerError)
```

Kotlin :

`NewInquirer` now can throw an exception


talwinder50 (approved):

Moopli (approved):
LGTM. We might want a followup that allows adding additional jsonld contexts to preload, either through build configuration or through mobile API.

DRK3:
Just a note for the breaking changes above: Kotlin also should handle error, so this needs to be documented as well. (don't need a code example, just need to let users of the SDK know to put the call in a try block if they're not already)

DRK3:
There's a typo in the commit title - Should be `VCs` not `VC's`.

cmd/wallet-sdk-gomobile/openid4ci/interaction_test.go
```diff
@@ -84,6 +84,21 @@ func TestNewInteraction(t *testing.T) {
...
 
+	t.Run("Success http timeout", func(t *testing.T) {
+		kms, err := localkms.NewKMS(localkms.NewMemKMSStore())
+		require.NoError(t, err)
+
+		resolver := &mockResolver{keyWriter: kms}
+
+		requiredArgs := openid4ci.NewArgs(createTestRequestURI("example.com"), kms.GetCrypto(), resolver)
+		opts := openid4ci.NewOpts()
+		opts.SetHTTPTimeoutNanoseconds(10000000)
```
DRK3, line 95: For readability, I suggest passing in an equivalent time value from the time package instead of `10000000`

demo/app/ios/Runner/Tests/IntegrationTest.swift
```diff
@@ -69,7 +69,7 @@ class IntegrationTest: XCTestCase {
         let vpInteraction = Openid4vpNewInteraction(openID4VPArgs, opts, nil)!
 
         let credentialsQuery = try vpInteraction.getQuery()
-        let inquirer = CredentialNewInquirer(nil)!
+        let inquirer = CredentialNewInquirer(nil, nil)!
```
DRK3, line 72: Don't ignore the error; need to check it

pkg/openid4vp/openid4vp.go
```diff
@@ -85,19 +85,7 @@ func New(
 	documentLoader ld.DocumentLoader,
 	opts ...Opt,
 ) *Interaction {
-	client, activityLogger, metricsLogger, networkDocumentLoaderHTTPTimeout := processOpts(opts)
-
-	if documentLoader == nil {
```
DRK3, line 90: What happens if no document loader is passed in now?

DRK3:
> Just a note for the breaking changes above: Kotlin also should handle error, so this needs to be documented as well. (don't need a code example, just need to let users of the SDK know to put the call in a try block if they're not already)

Also, on a similar note: the Android app's Kotlin code needs to add exception checks (there aren't any right now)

demo/app/ios/Runner/Tests/IntegrationTest.swift
vkubiv: We always ignore errors like this. So we need to handle all of them in tests or ignore them like previously.

DRK3:
> > Just a note for the breaking changes above: Kotlin also should handle error, so this needs to be documented as well. (don't need a code example, just need to let users of the SDK know to put the call in a try block if they're not already)
> 
> Also, on a similar note: the Android app's Kotlin test code needs to add exception checks (there aren't any right now)

Actually, nevermind about this - I suppose it doesn't really matter that much for the tests.

pkg/openid4vp/openid4vp.go
vkubiv: Created one with with embedded contexts

demo/app/ios/Runner/Tests/IntegrationTest.swift
DRK3: It's a mistake to ignore the errors. We need to at least start adding error checks for new code.

pkg/openid4vp/openid4vp.go
vkubiv: The default document loader is set on the go mobile part.

DRK3: The Go API is also public, though. It can be used independently of the gomobile API, so it needs some default behaviour too.

cmd/wallet-sdk-gomobile/openid4ci/interaction.go
```diff
@@ -235,9 +240,23 @@ func createGoAPIClientConfig(config *Args,
 		}
 
 		goAPIClientConfig.DocumentLoader = documentLoaderWrapper
+	} else {
+		dlHTTPClient := &http.Client{}
```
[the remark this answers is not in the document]
DRK3: I would say so. If I as a caller specified the DisableHTTPClientTLSVerification while using the default document loader and then got a TLS error during execution of the Wallet-SDK methods, I would think it's a bug. I guess we missed this before...

pkg/openid4vp/openid4vp.go
vkubiv: We pass a lot of parameters there are like didResolver, crypto, signatureVerifier that do not have default behavior too. We need some consistency. I am not sure that this is part of this optimization PR

cmd/wallet-sdk-gomobile/openid4ci/interaction.go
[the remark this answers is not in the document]
vkubiv: good

pkg/openid4vp/openid4vp.go
DRK3: I see what you mean, good point! Just wondering - was there a technical reason you removed the default behaviour that was already there, or was it just more for consistency? (Either way, it's OK by me)

demo/app/ios/Runner/Tests/IntegrationTest.swift
DRK3: Perhaps sometime in the future we can add in checks for the other existing errors (not in this PR of course)

pkg/openid4vp/openid4vp.go
DRK3: On a side note - we should technically be returning errors and/or have some default behaviour if those other required params aren't passed in...

vkubiv: I moved it to mobile wrappers, because for some code default document loader creation was already done on mobile wrappers. It should be consistent.

cmd/wallet-sdk-gomobile/openid4ci/interaction.go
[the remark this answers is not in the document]
vkubiv: done

cmd/wallet-sdk-gomobile/openid4ci/interaction_test.go
vkubiv: dene

cmd/wallet-sdk-gomobile/openid4ci/interaction_test.go
```diff
@@ -84,6 +84,21 @@ func TestNewInteraction(t *testing.T) {
 		require.NotNil(t, interaction)
 	})
 
+	t.Run("Success http timeout", func(t *testing.T) {
```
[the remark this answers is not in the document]
vkubiv: done

demo/app/ios/Runner/Tests/IntegrationTest.swift
vkubiv: done

DRK3:
The commit title message still says "VC's" - that should be fixed too.

demo/app/ios/Runner/Tests/IntegrationTest.swift
```diff
@@ -69,7 +69,9 @@ class IntegrationTest: XCTestCase {
         let vpInteraction = Openid4vpNewInteraction(openID4VPArgs, opts, nil)!
 
         let credentialsQuery = try vpInteraction.getQuery()
-        let inquirer = CredentialNewInquirer(nil)!
+
+        var newInquirerError: NSError?
```
DRK3, line 73: You're passing in an error now, but I don't see you checking it anywhere. So the error is still being ignored.

cmd/wallet-sdk-gomobile/docs/usage.md
```diff
@@ -1145,7 +1145,8 @@ let opts = Openid4vpNewOpts().setActivityLogger(activityLogger) // Optional, but
 var newInteractionError: NSError?
 let interaction = Openid4vpNewInteraction(args, opts, &newInteractionError)
 let query = interaction.getQuery()
-let inquirer = CredentialNewInquirer(nil)
+var newInquirerError: NSError?
+let inquirer = CredentialNewInquirer(nil,&newInquirerError)
```
DRK3, line 1149: Formatting: should have a space between the arguments

vkubiv: done

demo/app/ios/Runner/OpenID4VP.swift
```diff
@@ -53,7 +53,7 @@ public class OpenID4VP {
             throw OpenID4VPError.runtimeError("OpenID4VP interaction not properly initialized, call processAuthorizationRequest first")
         }
         
-            return try CredentialNewInquirer(CredentialInquirerOpts()?.setDIDResolver(didResolver))!.getSubmissionRequirements(vpQueryContent, credentials: storedCredentials)
+            return try CredentialNewInquirer(CredentialInquirerOpts()?.setDIDResolver(didResolver), nil)!.getSubmissionRequirements(vpQueryContent, credentials: storedCredentials)
```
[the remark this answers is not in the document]
vkubiv: I suggest to create a separate issue for all missing error handling in the app and fixing it there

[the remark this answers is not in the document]
DRK3: OK, sounds good :)

DRK3 (approved):

DRK3:
> The commit title message still says "VC's" - that should be fixed too.

To clarify: I'm talking about the actual Git commit title message (I see the PR title on GitHub has been fixed)

PR #467, opened 2023-07-04
test(app): IOS (swift) test for auth code flow

talwinder50:


rolsonquadras (approved):

demo/app/ios/Runner/Tests/IntegrationTest.swift
```diff
@@ -99,4 +100,95 @@ class IntegrationTest: XCTestCase {
...
+          scopes!.append("openid")!.append("profile")!
+
+          let opts = Openid4ciNewCreateAuthorizationURLOpts()!.setScopes(scopes!)
+          let authorizationLink = ciInteraction!.createAuthorizationURL("oidc4vc_client", redirectURI: "http://127.0.0.1/callback", opts: opts!, error: nil)
+          XCTAssertTrue(authorizationLink  != "", "authorizationLink:" + authorizationLink)
+
+           var redirectURL = ""
+           let r = Redirect()
+           let url = URL(string: authorizationLink)!
+           r.makeRequest(url: url, callback: { (location) in
```
vkubiv, line 142: Maybe it better to move that logic to some helper function. It will make test code more clear

vkubiv (approved):

demo/app/ios/Runner/Tests/IntegrationTest.swift
talwinder50: Aah i missed ur comment will do that in follow up pr

PR #470, opened 2023-07-05
docs(sdk): Update README.md for the wallet-sdk-gomobile

raphaelguye:
During the installation of `go`, I was stuck a bit because `gomobile` didn't worked.
[This issue](https://github.com/golang/go/issues/16935) helped me. 

I think it's worth modifying the README.

### DIFF
--- a/cmd/wallet-sdk-gomobile/README.md
+++ b/cmd/wallet-sdk-gomobile/README.md
@@ -6,11 +6,9 @@ This package contains the `gomobile`-compatible version of the SDK. It acts as a
 
 * [Go 1.20](https://go.dev/doc/install) or newer
 * The gomobile tools:
-  ```
-  go install golang.org/x/mobile/cmd/gomobile@latest
-  export PATH="$PATH:$HOME/go/bin"
-  gomobile init
-  ```
+  * `go install golang.org/x/mobile/cmd/gomobile@latest`
+  * `$GOPATH/bin` added to your path
+  * ``gomobile init``
 ### Android Bindings
 
 * Android SDK (installable via the SDK Manager in [Android Studio](https://developer.android.com/studio/install))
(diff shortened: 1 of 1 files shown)


raphaelguye:
DCO signed, ready to build again

raphaelguye:
title updated, ready to build again

DRK3:
Hi @raphaelguye, thanks for finding this issue and for making a pull request!

I'd like to suggest that instead of adding `export PATH="$PATH:$HOME/go/bin"` to the `The gomobile tools` section, instead what you can do is move the existing `$GOPATH/bin added to your path` bullet point above the `The gomobile tools` point.

The reasons why are because:
* The line you added is specific to Unix - it wouldn't work as expected on a Windows machine.
* While the GOPATH environment variable defaults to `$HOME/go`, it could be different on someone else's machine.
* The `export` command is just one way of adding GOPATH to your path, and once the terminal window is closed it'll be lost. Rather than picking a specific method of adding GOPATH to the path, we can simply tell the user to add it to their path and leave it up to them to choose how they want to do that.

By placing `$GOPATH/bin added to your path` above the `The gomobile tools` section, that tells the user that it's required before they can run those commands, which should help people in the future not run into the same issue you had, and the wording is agnostic to the user's OS + GOPATH env. var.

Let me know what you think :)

cmd/wallet-sdk-gomobile/README.md
```diff
@@ -6,12 +6,9 @@ This package contains the `gomobile`-compatible version of the SDK. It acts as a
...
 * The gomobile tools:
-  ```
-  go install golang.org/x/mobile/cmd/gomobile@latest
-  gomobile init
-  ```
-* `$GOPATH/bin` added to your path
-
+  * `go install golang.org/x/mobile/cmd/gomobile@latest`
+  * `$GOPATH/bin` added to your path
+  * ``gomobile init``
```
DRK3, line 11:
```suggestion
  * `gomobile init`
```

rolsonquadras:
Closing this as there is no response.

PR #482, opened 2023-07-11
feat(sdk)!: Improvements to dynamic client registration API

DRK3:
Breaking change: Changes to names and methods in dynamic client registration API. See below for details.

* In the Gomobile API, changed the scopes parameter type to a StringArray. Internally, it will be converted to the space-separated string that is required by the dynamic client registration spec, so this way the user of the SDK doesn't have to deal with splitting/parse spaces. Also, changed the method names from `scope` to `scopes`.
* In the Gomobile API, removed the hasClientMetadata method from the response object. Response objects don't actually have a nested clientMetadata object - the ClientMetadata object is just an abstraction here to make code cleaner and allow for reuse - so it doesn't make sense to ask if the object exists or not.
* For the response object, switched it from having a ClientMetadata object to a RegisteredMetadata object. While the objects fields are the same, the context and meaning of them is different, which justifies having different names.

### DIFF
--- a/cmd/wallet-sdk-gomobile/oauth2/registerclientresponse.go
+++ b/cmd/wallet-sdk-gomobile/oauth2/registerclientresponse.go
@@ -8,7 +8,9 @@ package oauth2
 
 import (
 	"errors"
+	"strings"
 
+	"github.com/trustbloc/wallet-sdk/cmd/wallet-sdk-gomobile/api"
 	goapioauth2 "github.com/trustbloc/wallet-sdk/pkg/oauth2"
 )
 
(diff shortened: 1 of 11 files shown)


cmd/wallet-sdk-gomobile/oauth2/registerclientresponse.go
```diff
@@ -62,18 +64,106 @@ func (r *RegisterClientResponse) ClientSecretExpiresAt() (int, error) {
 	return *r.goAPIRegisterClientResponse.ClientSecretExpiresAt, nil
 }
 
-// HasClientMetadata indicates whether this RegisterClientResponse has client metadata.
-func (r *RegisterClientResponse) HasClientMetadata() bool {
-	return r.goAPIRegisterClientResponse.ClientMetadata != nil
+// RegisteredMetadata returns the RegisteredMetadata object, which can be used to determine what metadata was
```
talwinder50, line 67: How the registeredMetadata is different from the clientMetadata ?

rolsonquadras (approved):

cmd/wallet-sdk-gomobile/oauth2/registerclientresponse.go
DRK3: It's the metadata that actually gets registered on the server - the spec states that the server may reject or replace the client metadata values. The client should know about this so they can react accordingly (if necessary). See here for more info: https://datatracker.ietf.org/doc/html/rfc7591#section-3.2.1

DRK3: So using the the existing name (client metadata) in the response is misleading, since it may not match the client metadata that was sent. Otherwise, the fields are the same (except for the issuer state field, but that will be removed soon as it was just a temporary workaround)

vkubiv (approved):

PR #487, opened 2023-07-14
feat(sdk): openid4ci js api.

vkubiv:


birtony (approved):

cmd/wallet-sdk-js/jsinterop/indexeddb/indexeddb.go
```diff
@@ -0,0 +1,658 @@
...
+	"fmt"
+	"strings"
+	"sync"
+	"syscall/js"
+	"time"
+
+	"github.com/hyperledger/aries-framework-go/spi/storage"
+)
+
+// TODO (#2528): Proper implementation of all methods.
```
DRK3, line 25: I think you should remove the extra TODOs and methods that aren't applicable here (there's a bunch of things done here for aries-specific things). (At a minimum, the TODOs should be removed at least to avoid confusion in the future)

cmd/wallet-sdk-js/walletsdk/agent.go
```diff
@@ -81,7 +95,48 @@ func (a *Agent) CreateOpenID4CIIssuerInitiatedInteraction(
 	}
 
 	return &OpenID4CIIssuerInitiatedInteraction{
-		interaction: interaction,
+		Interaction: interaction,
 		crypto:      a.crypto,
 	}, nil
 }
+
+// ResolveDisplayData resolves credential display data in openid4ci Interaction.
```
DRK3, line 103:
```suggestion
// ResolveDisplayData resolves display information for issued credentials based on an issuer's metadata, which is fetched
// using the issuer's (base) URI.
// The CredentialDisplays in the returned Data object correspond to the VCs passed in and are in the
// same order.
```

cmd/wallet-sdk-js/jsinterop/indexeddb/indexeddb.go
vkubiv: Storage will be injectable from JS. And will be implemented in JS too.

cmd/wallet-sdk-js/walletsdk/agent.go
```diff
@@ -81,7 +95,48 @@ func (a *Agent) CreateOpenID4CIIssuerInitiatedInteraction(
...
+		credentialschema.WithCredentials(parsedCreds),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("resolve data: %w", err)
+	}
+
+	return data, nil
+}
+
+// ParseCredential parse credential.
```
DRK3, line 131:
```suggestion
// ParseCredential parses the given serialized VC into a VC object.
```

cmd/wallet-sdk-js/jsinterop/indexeddb/indexeddb.go
DRK3: You should still at least get rid of the TODOs that don't make sense in this project, and any unused public methods. It's better to not have it in the commit history.

rolsonquadras (approved):

DRK3 (changes requested):
See comments

rolsonquadras:
The requested changes will be addressed in next PR.

DRK3:
> The requested changes will be addressed in next PR.

That's OK, don't have to make a PR just for this (can be done when the storage is reworked) :)

PR #490, opened 2023-07-17
feat(sdk): add OpenID4VP present API that skips constraint validation

Moopli:


pkg/openid4vp/openid4vp.go
```diff
@@ -148,8 +148,31 @@ func (o *Interaction) VerifierDisplayData() (*VerifierDisplayData, error) {
...
 // PresentCredential presents credentials to redirect uri from request object.
 func (o *Interaction) PresentCredential(credentials []*verifiable.Credential) error {
+	return o.presentCredentials(
+		credentials,
+		&presentOpts{},
+	)
+}
+
+// PresentCredentialUnsafe presents credentials to redirect uri from request object.
+// If one credential is provided, this will skip presentation definition constraint validation.
```
DRK3, line 164: Just to clarify: do you mean that the presentation definition constraint validation is only skipped if one (and only one) credential is provided? I noticed that the Gomobile version's comment doesn't state this.

Moopli: Good point, I'll clarify.

Moopli: Actually, I should also add a check that fails if more than one VC is passed into unsafe, since that shouldn't work.

DRK3: You can just make it accept only one credential instead of an array.

pkg/openid4vp/openid4vp.go
```diff
@@ -148,8 +148,31 @@ func (o *Interaction) VerifierDisplayData() (*VerifierDisplayData, error) {
...
 func (o *Interaction) PresentCredential(credentials []*verifiable.Credential) error {
+	return o.presentCredentials(
+		credentials,
+		&presentOpts{},
+	)
+}
+
+// PresentCredentialUnsafe presents a single credential to redirect uri from request object.
+// This skips presentation definition constraint validation.
+func (o *Interaction) PresentCredentialUnsafe(credentials []*verifiable.Credential) error {
```
DRK3, line 165: Switch it from an array of credentials to a single credential and then you can remove the multiple credential check

cmd/wallet-sdk-gomobile/openid4vp/interaction.go
```diff
@@ -159,6 +160,17 @@ func (o *Interaction) PresentCredential(credentials *verifiable.CredentialsArray
...
 }
 
+// PresentCredentialUnsafe presents a single credential to redirect uri from
+// request object.
+//
+// Note: this variant of PresentCredential will skip client-side presentation
+// definition constraint validation. All input descriptors will accept the
+// provided credential, at least in terms of issuer fields, and subject data
+// fields.
+func (o *Interaction) PresentCredentialUnsafe(cred *verifiable.Credential) error {
```
DRK3, line 170: For consistency with the `PresentCredential` method, this argument should probably be `credential` instead of `cred`. (Sometimes the objective-C bindings require the argument name to be passed in, which means a breaking change if we want to make the naming consistent later)

DRK3 (approved):
Left one more minor comment

cmd/wallet-sdk-gomobile/openid4vp/interaction.go
Moopli: Had to rename it to avoid shadowing the import, would the API look cleaner if I renamed the import instead?

DRK3: From the caller's point of view, it would be slightly cleaner since the argument names are a bit more consistent. But I admit this is a very, very small thing...

DRK3: But I see your point - the import renaming might be a bit ugly too, so... go with your gut :D

rolsonquadras (approved):

vkubiv (approved):

cmd/wallet-sdk-gomobile/openid4vp/interaction.go
Moopli: Did the change, I think you're right :)

pkg/openid4vp/openid4vp.go
Moopli: Done

pkg/openid4vp/openid4vp.go
Moopli: Done

PR #491, opened 2023-07-17
feat(sdk): Error improvements

DRK3:
* The error response structure defined in the OpenID4CI spec is now parsed, and the error codes defined in the spec are now passed through to the caller (with newly-defined Wallet-SDK error codes).
* Fixed some places where error codes were wrapping lower-level error codes. This wrapping would cause the error parsing package to only "see" the higher-level error, which isn't as helpful for the user.
* Defined a new common error code for invalid usage of Wallet-SDK. Removed specific invalid usage error codes to help prevent callers from using those error code for unintended purposes (e.g. using the PIN_REQUIRED error code instead of checking if a PIN is required using the proper Wallet-SDK method).
* Fixed Gomobile API methods that weren't converting the Go API wallet error types to their gomobile equivalents.
* Added a Full field to the parsed gomobile error type. The ToMobileError function parses the first Go API Wallet Error type it finds in the error chain into the Code, Category, and Details fields. If there were any higher-level errors above that, they get lost. Now, if these higher-level error details will be captured in the Full field.
* Improved some error code names.
* Updated documentation.
* Switched OpenID4CI error codes from iota to manual numbering to make them easier to maintain.


rolsonquadras (approved):

cmd/wallet-sdk-gomobile/wrapper/error.go
```diff
@@ -42,6 +42,7 @@ func ToMobileErrorWithTrace(err error, trace *otel.Trace) error {
 			Category: walletError.Scenario,
 			Details:  walletError.ParentError,
 			TraceID:  traceID,
+			Raw:      err.Error(),
```
vkubiv, line 45: 'Raw' name is a little bit unclear. And also it looks like invalid usage of our errors model,  walleterror.Error shouldn't be wrapped just with a simple error. Original walleterror.Error should be handled, and new walleterror.Error produced.

vkubiv (changes requested):

cmd/wallet-sdk-gomobile/wrapper/error.go
vkubiv: In case when a top-level error is not walleterror.Error, maybe it will be good to treat this as an error, or even panic. Because this means that we just done an error in the wallet-sdk code.

pkg/openid4ci/errors.go
```diff
@@ -6,41 +6,51 @@ SPDX-License-Identifier: Apache-2.0
...
 )
 
-// Constants' names and reasons are obvious so they do not require additional comments.
+// Constants' names and reasons are obvious, so they do not require additional comments.
 // nolint:golint,nolintlint
 const (
-	NoClientConfigProvidedCode = iota
-	ClientConfigNoDIDResolverProvidedCode
-	InvalidIssuanceURICode
+	InvalidIssuanceURICode = iota
```
vkubiv, line 37: This is not related to your PR, but maybe we need to assign direct value to error codes, so that when we rearrange or delete them they are not changing values. We can bring this into discussion in our daily

rolsonquadras: I agree with this. It's better to have direct values as it would be easier to see in the PR diff if it's changed by mistake. It's hard to find such things with iota.

cmd/wallet-sdk-gomobile/wrapper/error.go
DRK3: I see some challenges here...

Many of the error codes that come from Wallet-SDK happen at a low-level (i.e. in nested functions). In Go, it's quite beneficial to wrap errors from lower-level functions, which helps provide more information about the source of an error for debugging purposes. So, if I were to only pass through the low-level error without wrapping, we're losing potentially quite a lot of contextual information. Also, keep in mind that with the new use-cases and flows we're supporting (e.g. Wallet-initiated flow, pre-auth, auth, different authorization schemes), theres more and more places where the same low-level functions are called from higher-level functions. So, there will be a lot of places where we have to ensure we never wrap errors (and lots of good context we're going to lose).

With that in mind, this will be challenging to maintain, as this becomes an "implicit" restriction of Wallet-SDK development, which will be difficult to maintain. Note that with our current error code structure, we've already introduced one implicit restriction - where you need to ensure that you never wrap a walleterror.Error with another walleterror.Error, since that causes the lower-level walleterror.Error to be "unseen" by the gomobile parse function... we should keep these implicit restrictions to a minimum.

So, my idea with this "Raw" field was to try to get the best of both worlds - allow for error code checks by the calling code, while also providing a way to ensure that extra higher-level contextual error messages can be seen as well, if desired.

Do you think there's another way we can achieve this "best of both worlds" solution I'm looking for, without minimizing implicit restrictions on how errors are handled in Go code?

P.S. Perhaps "Full" would be a better name than "Raw"?

DRK3:
> about this part: "Removed some error codes that were reporting issues related to incorrect usage of Wallet-SDK. These error codes don't really serve any purpose, as these situations should only really exist during development and the end-user of an application using Wallet-SDK won't encounter them."
> 
> Developers that use our SDK will be confused in this case. I think it is better to add an additional walletError type. We have runtime and validation errors, we can add something logic error type, that will indicate improper use of SDK. This error type can have the same error code, and error category for all of them. In this case, users of our SDK can uniformly process all errors.

Note that I still return errors when the SDK is used incorrectly - the only difference now is that I'm not defining an explicit error code. There's several reasons why I believe it's better to not return an explicit error code in these cases:

1. Regardless of how they handle the errors from Wallet-SDK (whether they actually parse them, or they simply print out the JSON error structure), the underlying error message will be passed through, as the existing gomobile error handling/parsing code ensures that even errors that aren't using the walleterror.Error type will be passed through, just with a generic error code instead.
2. From what I can see, there are two main benefits to having explicit error codes are: the first is to allow the caller to detect certain errors programatically and allow them to handle them in a special way (e.g. by doing retries in some cases). The second is to allow an application an easy way to provide a short, concise error to an end-user that they can then submit in a bug report or to something like customer support, without the huge underlying error that could be confusing. In cases where the error is due to improper usage of Wallet-SDK (e.g. missing arguments, calling the wrong method, etc), neither of these cases are applicable. There's no action a caller can take can take to programatically respond to it (they just have to fix the code, perhaps by reading the docs), and the end-user should never see these error codes.
3. We have many, many, many places where we return errors due to improper arguments (e.g. nil args). It would be very difficult to add an error code for each one of these scenarios, and would bloat the code considerably.
4. For some of the improper usage scenarios, the error codes may cause the caller to "fix" them in an incorrect way. For example - one of the error codes I removed was the "PIN_REQUIRED" one. This error happens only if the caller doesn't check to see if a PIN is required first, which they should do using the `pinRequired` method from the SDK itself. If I return an explicit error code, that may lead to the caller using that as their mechanism to determining whether the PIN is required, which isn't the proper way to do it. So it's better to not make it easy for them to do the wrong way.

vkubiv:
> > about this part: "Removed some error codes that were reporting issues related to incorrect usage of Wallet-SDK. These error codes don't really serve any purpose, as these situations should only really exist during development and the end-user of an application using Wallet-SDK won't encounter them."
> > Developers that use our SDK will be confused in this case. I think it is better to add an additional walletError type. We have runtime and validation errors, we can add something logic error type, that will indicate improper use of SDK. This error type can have the same error code, and error category for all of them. In this case, users of our SDK can uniformly process all errors.
> 
> Note that I still return errors when the SDK is used incorrectly - the only difference now is that I'm not defining an explicit error code. There's several reasons why I believe it's better to not return an explicit error code in these cases:
> 
> 1. Regardless of how they handle the errors from Wallet-SDK (whether they actually parse them, or they simply print out the JSON error structure), the underlying error message will be passed through, as the existing gomobile error handling/parsing code ensures that even errors that aren't using the walleterror.Error type will be passed through, just with a generic error code instead.
> 2. From what I can see, there are two main benefits to having explicit error codes are: the first is to allow the caller to detect certain errors programatically and allow them to handle them in a special way (e.g. by doing retries in some cases). The second is to allow an application an easy way to provide a short, concise error to an end-user that they can then submit in a bug report or to something like customer support, without the huge underlying error that could be confusing. In cases where the error is due to improper usage of Wallet-SDK (e.g. missing arguments, calling the wrong method, etc), neither of these cases are applicable. There's no action a caller can take can take to programatically respond to it (they just have to fix the code, perhaps by reading the docs), and the end-user should never see these error codes.
> 3. We have many, many, many places where we return errors due to improper arguments (e.g. nil args). It would be very difficult to add an error code for each one of these scenarios, and would bloat the code considerably.
> 4. For some of the improper usage scenarios, the error codes may cause the caller to "fix" them in an incorrect way. For example - one of the error codes I removed was the "PIN_REQUIRED" one. This error happens only if the caller doesn't check to see if a PIN is required first, which they should do using the `pinRequired` method from the SDK itself. If I return an explicit error code, that may lead to the caller using that as their mechanism to determining whether the PIN is required, which isn't the proper way to do it. So it's better to not make it easy for them to do the wrong way.

about 3. We can just have one error for all improper usage. So it will be something like this NewImproperUsage(err error). In the end, it will translate into { "module": "general", "code": "0",  "category": "IMPROPER_USAGE", details: "..."  }. Because currently, we get UKN2-000 code and UNEXPECTED_ERROR category. It is not very clear 

I will approve you PR. If you like this idea you can implement it.

vkubiv (approved):

DRK3:
I've marked it WIP to prevent merging for now

cmd/wallet-sdk-gomobile/wrapper/error.go
vkubiv: Yes, "Full" would be better than Raw

vkubiv: In this case, only one implicit restriction is left, which is always return walleterror.Error from gomobile functions that are accessible from mobile code. We can also make this restriction a little bit stronger. By creating helper functions for tests, that will not only check that error has the correct message but also that it is walleterror.Error, with correct Code, and Category

vkubiv: Just realized here one other case. When we don't want to change the nature of the error, but just add additional trace information, so we know that it happened when call from a particular function. As I understand you mean this case. 
in this case, we can replace fmt.Errorf with a more clever function. That will keep the original walleterror.Error a top, but just add additional info at the beginning of the details filed of that original error.

DRK3: > Just realized here one other case. When we don't want to change the nature of the error, but just add additional trace information, so we know that it happened when call from a particular function. As I understand you mean this case. in this case, we can replace fmt.Errorf with a more clever function. That will keep the original walleterror.Error a top, but just add additional info at the beginning of the details filed of that original error.

Ah, I think I understand what you're saying. Let me look into this and see...

> I this it is good that gomobile parse function will not see lower-level walleterror.Error, but just see high-level walleterror.Error. In this case, "details" field will contain low-level error descriptions. So it can be used to debug error couses. WalletSDK users should react to higher-level errors. Low-level error in this case is something that depends on implementation and can be changed with implementation change.

I think based on your most recent comment, you might have realized this already, but just in case: it's actually the higher-level error details that were getting lost (rather than lower-level ones). Actually, all the lower-level details do get sent - if we want to hide them, that would be a different thing....

DRK3: > Do you think there's another way we can achieve this "best of both worlds" solution I'm looking for, without minimizing implicit restrictions on how errors are handled in Go code?

Oops... meant to say _while_ minimizing implicit restrictions, not _without_.

pkg/walleterror/walleterror.go
```diff
@@ -49,6 +50,16 @@ func NewSystemError(module string, code int, errorName string, parentError error
 	}
 }
 
+// NewInvalidSDKUsageError creates a new invalid SDK usage error, used to indicate when a Wallet-SDK API is used
+// incorrectly.
+func NewInvalidSDKUsageError(module string, parentError error) *Error {
+	return &Error{
+		Code:        getErrorCode(module, incorrectUsageError, 0),
```
vkubiv, line 57: I think it should be some big value like 10000 not 0, to not overlap with codes from the particular module.

cmd/wallet-sdk-gomobile/wrapper/error.go
DRK3: I'm thinking I can do this in a followup PR, since this one is already pretty big. Does that sound ok?

pkg/walleterror/walleterror.go
DRK3: The error codes are relative to the category (incorrect usage), so overlap shouldn't be possible, right?

DRK3: i.e. they will always start with OCI3-

PR #497, opened 2023-07-19
fix(sdk): fixed wallet-sdk-js publishing script

birtony:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -247,7 +247,7 @@ jobs:
           VERSION=$(git rev-parse --short=7 HEAD)
           echo "repoTag=${TAG} repoVersion=${VERSION}"
           if [[ $TAG == $VERSION ]]; then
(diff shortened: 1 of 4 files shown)


talwinder50 (approved):

.github/workflows/build.yml
```diff
@@ -431,19 +431,14 @@ jobs:
           echo $PATH
           
           cd ./cmd/wallet-sdk-js
-          sed -i 's/"version": "0.1.0"/"version": "'${ env.NEW_VERSION }'"/g' package.json
+          sed -i 's/"version": "0.1.0"/"version": "'${{ env.NEW_VERSION }}'"/g' package.json
```
talwinder50, line 434: Why we needed double brackets here?

birtony: We need them to use the actual value of the `NEW_VERSION` environment variable that's set in the previous step

PR #500, opened 2023-07-20
fix(app): removed ffi lib to unblock web app

birtony:
`dart:ffi` is specifically designed for native code interoperability in Flutter mobile apps, but does not work in web. Need to find a solution to avoid break the web app.


talwinder50 (approved):

talwinder50:
I will look through this to make it work in separate pr

rolsonquadras (approved):

PR #508, opened 2023-07-24
feat(sdk): OpenID4VP error response parsing

DRK3:
* Added parsing for OpenID4VP error responses per the spec + new error codes.
*  For consistency, replaced one of the existing OpenID4VP error codes with the new "invalid SDK usage" error code type introduced in the OpenID4CI error handling commit.
* Previously, when errors without error type info were consumed in an app using the gomobile bindings, the default error code and category would differ depending on which layer the error processing was done (would be either GNR2-000/GENERAL_ERROR or UKN2-000/UNEXPECTED_ERROR). I've updated it so that we use the same error code and category in both places. I chose "OTHER_ERROR" for the category to try to be more descriptive/accurate of what the category is.

### DIFF
--- a/cmd/wallet-sdk-gomobile/walleterror/error.go
+++ b/cmd/wallet-sdk-gomobile/walleterror/error.go
@@ -26,8 +26,8 @@ func Parse(message string) *Error {
 	err := json.Unmarshal([]byte(message), walletErr)
 	if err != nil {
 		return &Error{
-			Code:     "GNR2-000",
-			Category: "GENERAL_ERROR",
+			Code:     "UKN2-000",
+			Category: "OTHER_ERROR",
 			Details:  message,
 		}
 	}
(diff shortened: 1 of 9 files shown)


cmd/wallet-sdk-gomobile/walleterror/error.go
```diff
@@ -26,8 +26,8 @@ func Parse(message string) *Error {
 	err := json.Unmarshal([]byte(message), walletErr)
 	if err != nil {
 		return &Error{
-			Code:     "GNR2-000",
-			Category: "GENERAL_ERROR",
+			Code:     "UKN2-000",
+			Category: "OTHER_ERROR",
```
talwinder50, line 30: Why did we change the category and code here, does it impact the issuance flow error ?

DRK3: Check the commit message for the reason behind this change. This error type only gets returned when there's no underlying error code - we should never be checking for this error code from the app. Note: I do see one place in handle_redirect_uri.dart where the `UKN2-000` is checked, but this appears to be a mistake and I don't think error handling works there as expected. We should fix this when we overhaul the demo app's error handling @rolsonquadras @talwinder50

rolsonquadras (approved):

talwinder50 (approved):

Issue #510, opened 2023-07-25
Make the button submit/ Cancel button fix to the bottom 


talwinder50:
right now they drag further to the Botton  when error widget appears

PR #511, opened 2023-07-25
feat(app): Update the invalid pin error handling scenario

talwinder50:
- Used wallet sdk parsed error api to get the parsed details and category 
- Based on  parsed Category (INVALID_GRANT) a messaging is shown to the user **"Oops! Something went wrong! 
Try re-entering the pin or scan the new QR code."** 
- For the other categories we will show what ever is been sent in the details by parsed method. 
- If user has entered wrong pin they can try re-entering or restart the operation as per the messaging. 
   - If pin was wrong and user re-entered the right pin, user will be redirected to credential preview screen 

Things to fix: 
- Add a loader once the user hit the submit or renter button #509 
- Make the button submit/ Cancel button fix to the bottom right now they drag further to the Botton  when error widget appears #510
invalid_pin_error
invalid_pin_error_handling

### DIFF
--- a/demo/app/android/app/src/main/kotlin/walletsdk/flutter/app/MainActivity.kt
+++ b/demo/app/android/app/src/main/kotlin/walletsdk/flutter/app/MainActivity.kt
@@ -118,14 +118,6 @@ class MainActivity : FlutterActivity() {
                             }
                         }
 
-                        "fetchDID" -> {
-                            try {
-                                val didID = call.argument<String>("didID")
-                            } catch (e: Exception) {
-                                result.error("Exception", "Error while setting fetched DID", e)
-                            }
-                        }
-
                         "serializeDisplayData" -> {
                             try {
                                 val credentialDisplay = serializeDisplayData(call)
@@ -161,6 +153,15 @@ class MainActivity : FlutterActivity() {
                                 result.success(issuerURIResp)
                             } catch (e: Exception) {
                                 result.error("Exception", "Error while getting issuerURI", e)
+                            }
+                        }
+
+                        "parseWalletError" -> {
+                            try {
+                                val parsedWalletError = parseWalletSDKError(call)
+                                result.success(parsedWalletError)
+                            } catch (e: Exception) {
+                                result.error("Exception", "Error while parsing wallet sdk error", e)
                             }
                         }
 
@@ -400,7 +401,19 @@ class MainActivity : FlutterActivity() {
     }
 
 
+    private fun parseWalletSDKError(call: MethodCall): MutableMap<String, String> {
+        val localizedErrorMessage = call.argument<String>("localizedErrorMessage") ?: throw java.lang.Exception("localizedErrorMessage is missing")
 
+        val parsedError = Walleterror.parse(localizedErrorMessage)
+
+        val parsedErrResp: MutableMap<String, String> = mutableMapOf()
+        parsedErrResp["category"] = parsedError.category
+        parsedErrResp["details"] = parsedError.details
+        parsedErrResp["code"] = parsedError.code
+        parsedErrResp["traceID"] = parsedError.traceID
+
+        return parsedErrResp
+    }
     /**
      * ResolveDisplay resolves display information for issued credentials based on an issuer's metadata, which is fetched
     using the issuer's (base) URI. The CredentialDisplays returns DisplayData object correspond to the VCs passed in and are in the
--- a/demo/app/ios/Runner/flutterPlugin.swift
+++ b/demo/app/ios/Runner/flutterPlugin.swift
@@ -43,6 +43,10 @@ public class SwiftWalletSDKPlugin: NSObject, FlutterPlugin {
             let otp = fetchArgsKeyValue(call, key: "otp")
             requestCredential(otp: otp!, result: result)
             
+        case "parseWalletSDKError":
+            let localizedErrorMessage = fetchArgsKeyValue(call, key: "localizedErrorMessage")
+            parseWalletSDKError(localizedErrorMessage: localizedErrorMessage!, result: result)
+            
         case "requestCredentialWithAuth":
             let redirectURIWithParams = fetchArgsKeyValue(call, key: "redirectURIWithParams")
             requestCredentialWithAuth(redirectURIWithParams: redirectURIWithParams!, result: result)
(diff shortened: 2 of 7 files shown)


rolsonquadras (approved):

demo/app/lib/views/otp.dart
```diff
@@ -150,28 +185,36 @@ class _OTPPage extends State<OTP> {
...
                         } catch (err) {
                           String errorMessage = err.toString();
+                          log("errorMessage-> $errorMessage");
                           if (err is PlatformException &&
                               err.message != null &&
                               err.message!.isNotEmpty) {
-                            errorMessage = err.details!;
+                            var resp =  await WalletSDKPlugin.parseWalletSDKError(localizedErrorMessage: err.details);
+                            if(resp.category == "INVALID_GRANT"){
+                              errorMessage = "Try re-entering the pin or scan the new QR code";
```
DRK3, line 194:
```suggestion
                              errorMessage = "Try re-entering the pin or scan a new QR code";
```

demo/app/ios/Runner/flutterPlugin.swift
```diff
@@ -528,12 +531,32 @@ public class SwiftWalletSDKPlugin: NSObject, FlutterPlugin {
...
+                                               details: parsedError))
+              
             }
     }
     
+    public func parseWalletSDKError(arguments: Dictionary<String, Any>, result: @escaping FlutterResult){
+        guard let localizedErrorMessage = arguments["localizedErrorMessage"] as? String else{
+            return  result(FlutterError.init(code: "NATIVE_ERR",
+                                             message: "error while parsing wallet sdk error",
+                                             details: "parameter localizedErrorMessage is missed"))
```
DRK3, line 546: I think you mean `localizedErrorMessage is missing`?

demo/app/ios/Runner/flutterPlugin.swift
```diff
@@ -528,12 +531,32 @@ public class SwiftWalletSDKPlugin: NSObject, FlutterPlugin {
...
-                                         details: error.localizedDescription))
+                                               details: parsedError))
+              
             }
     }
     
+    public func parseWalletSDKError(arguments: Dictionary<String, Any>, result: @escaping FlutterResult){
+        guard let localizedErrorMessage = arguments["localizedErrorMessage"] as? String else{
+            return  result(FlutterError.init(code: "NATIVE_ERR",
+                                             message: "error while parsing wallet sdk error",
```
DRK3, line 545:
```suggestion
                                             message: "error while parsing Wallet-SDK error",
```

demo/app/ios/Runner/flutterPlugin.swift
```diff
@@ -528,12 +531,32 @@ public class SwiftWalletSDKPlugin: NSObject, FlutterPlugin {
...
+        let parsedError = WalleterrorParse(localizedErrorMessage)
+        
+        var parsedErrorResult :[String:Any] = [
+            "category":   parsedError!.category,
+            "details":  parsedError!.details,
+            "code":  parsedError!.code,
+            "traceID":  parsedError!.traceID
+        ]
+        result(parsedErrorResult)
+    }
```
DRK3, line 557: Can you do the force unwrap once on line 548 instead of on every line from 551-554? Like this:

`let parsedError = WalleterrorParse(localizedErrorMessage)!`

demo/app/lib/views/otp.dart
```diff
@@ -150,28 +185,36 @@ class _OTPPage extends State<OTP> {
                           _navigateToCredPreviewScreen(credentials, issuerURI, serializeDisplayData!, userDIDId, credID);
                         } catch (err) {
                           String errorMessage = err.toString();
+                          log("errorMessage-> $errorMessage");
                           if (err is PlatformException &&
                               err.message != null &&
                               err.message!.isNotEmpty) {
-                            errorMessage = err.details!;
+                            var resp =  await WalletSDKPlugin.parseWalletSDKError(localizedErrorMessage: err.details);
+                            if(resp.category == "INVALID_GRANT"){
```
DRK3, line 193:
```suggestion
                            if (resp.category == "INVALID_GRANT"){
```

DRK3, line 193: There's one more error category you can add a check for: `INVALID_TOKEN`. In the case of `INVALID_TOKEN`, however, re-entering the PIN won't help, so in that case the error message should just tell the user to scan a new QR code.

DRK3 (commented):
Looks good overall - I suggested some recommended UI text changes and another error code we can check for to improve the user experience, however I don't see any major issues.

demo/app/android/app/src/main/kotlin/walletsdk/flutter/app/MainActivity.kt
```diff
@@ -400,7 +401,19 @@ class MainActivity : FlutterActivity() {
     }
 
 
+    private fun parseWalletSDKError(call: MethodCall): MutableMap<String, Any> {
+        val localizedErrorMessage = call.argument<String>("localizedErrorMessage") ?: throw java.lang.Exception("localizedErrorMessage is missed")
 
+        val parsedError = Walleterror.parse(localizedErrorMessage)
+
+        val parsedErrResp: MutableMap<String, Any> = mutableMapOf()
```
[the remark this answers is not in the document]
talwinder50: updated

demo/app/lib/wallet_sdk/wallet_sdk_mobile.dart
```diff
@@ -65,6 +66,12 @@ class WalletSDK extends WalletPlatform {
     }
   }
 
+  Future<WalletSDKError> parseWalletSDKError({required String localizedErrorMessage}) async {
+    var parsedWalletError = await methodChannel.invokeMethod(
+        'parseWalletSDKError', <String, dynamic>{'localizedErrorMessage': localizedErrorMessage});
```
DRK3, line 71: Can this map be made more specific?

demo/app/lib/wallet_sdk/wallet_sdk_model.dart
```diff
@@ -95,6 +95,45 @@ class SupportedCredentials {
...
+    required this.traceID,
+  });
+
+  @override
+  String toString() {
+    return 'WalletSDKError { code: $code, category: $category, details: $details, traceID: $traceID }';
+  }
+
+  Map<String, dynamic> toJson() {
+    final Map<String, dynamic> data = <String, dynamic>{};
```
DRK3, line 117: This map looks more generic than it needs to be

demo/app/lib/wallet_sdk/wallet_sdk_model.dart
```diff
@@ -95,6 +95,45 @@ class SupportedCredentials {
...
+    data['code'] = code;
+    data['category'] = category;
+    data['details'] = details;
+    data['traceID'] = traceID;
+    return data;
+  }
+
+  factory WalletSDKError.fromJson(Map<String, dynamic> json) {
```
DRK3, line 125: Same here

demo/app/lib/wallet_sdk/wallet_sdk_mobile.dart
talwinder50: You mean map of string , string ? Instead of dynamic 

Other than that walletsdkerror is the object type we are rendering the object which we get from sdk. 

I believe, It cant be more specific than that. I will change that if thats what u mean

DRK3: Even better: see if it can just be a simple string instead of a map

DRK3:
The commit says "iOS", but I see android code there too. Does this also add support for the Android app, or is that going to be a separate PR?

DRK3 (approved):

PR #515, opened 2023-07-27
feat(sdk): Improved authorization request error handling

DRK3:
* Improved error handling for the OpenID4VP authorization request.
* Make the validation and processing of the authorization request more robust - now the request_uri parameter can be URL encoded or not and it'll still work, and also it no longer has to be the first query parameter in the URL string in order for parsing to work correctly.
* As a result of the refactoring, the number of public interaction methods that have an error return have been reduced, which should help simplify calling code.


cmd/wallet-sdk-gomobile/openid4vp/interaction.go
```diff
@@ -146,13 +148,10 @@ func (o *Interaction) GetQuery() ([]byte, error) {
 }
 
 // VerifierDisplayData returns display information about verifier.
-func (o *Interaction) VerifierDisplayData() (*VerifierDisplayData, error) {
-	displayData, err := o.goAPIOpenID4VP.VerifierDisplayData()
-	if err != nil {
-		return nil, wrapper.ToMobileErrorWithTrace(err, o.oTel)
-	}
+func (o *Interaction) VerifierDisplayData() *VerifierDisplayData {
```
vkubiv, line 151: This is breaking change. You can keep old signature with error '(*VerifierDisplayData, error)'. Just always return nil. Or add ! to PR title.

vkubiv (changes requested):

cmd/wallet-sdk-gomobile/openid4vp/interaction.go
DRK3: It doesn't actually break any existing code (just an IDE warning/complaint for iOS)

vkubiv (approved):

rolsonquadras (approved):

Moopli (approved):

PR #517, opened 2023-07-31
chore(app): updated web targets in demo app and sdk version (WIP)

birtony:


demo/app/lib/wallet_sdk/wallet_sdk_js.dart
```diff
@@ -177,6 +177,10 @@ class WalletSDK extends WalletPlatform {
     }
   }
 
+  Future<WalletSDKError> parseWalletSDKError({required String localizedErrorMessage}) async {
+    return WalletSDKError.fromJson({});
```
vkubiv, line 181: I think it is better to throw an error, something like `throw Exception('Method not implemented')`

vkubiv (changes requested):

vkubiv (approved):

talwinder50 (approved):

PR #539, opened 2023-08-08
fix(app): Add the verfier logo uri and purpose in the UI

talwinder50:


rolsonquadras (approved):

demo/app/lib/views/presentation_preview.dart
```diff
@@ -30,6 +30,8 @@ class PresentationPreviewState extends State<PresentationPreview> {
   var uuid = const Uuid();
 
   late String verifierName = '';
+  String? verifierLogoURL;
```
DRK3, line 33: Keep in mind that Wallet-SDK doesn't return an optional string - if no logo URL is present, then it'll be a blank string. You'll have to map the blank string to a null `String?` or, instead, make this a non-optional String and check for an empty string below on line 69 instead. (this also applies to the other files as well)

DRK3 (approved):
Looks good, just keep in mind the comment I left regarding `verifierLogoURL`

PR #551, opened 2023-08-14
feat(app): Update supported credential UI in the wallet initiated flow

talwinder50:

### DIFF
--- a/demo/app/lib/views/supported_credentials_list.dart
+++ b/demo/app/lib/views/supported_credentials_list.dart
@@ -4,7 +4,7 @@ Copyright Gen Digital Inc. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 import 'package:app/models/connect_issuer_config_value.dart';
-import 'package:app/widgets/primary_button.dart';
+import 'package:cached_network_image/cached_network_image.dart';
 import 'package:flutter/material.dart';
 
 import 'package:app/wallet_sdk/wallet_sdk_mobile.dart';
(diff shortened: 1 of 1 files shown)


demo/app/lib/views/supported_credentials_list.dart
```diff
@@ -51,31 +51,47 @@ class SupportedCredentialsListState extends State<SupportedCredentialsList> {
...
-                      icon: const Icon(Icons.credit_card_sharp),
-                      tooltip: 'Credential',
-                      onPressed: () {},
+                    title: Text(
+                      widget.supportedCredentialList.elementAt(index).display.elementAt(index).name,
+                      style: TextStyle(
+                        fontSize:14,
+                        fontWeight: FontWeight.bold,
+                        color: '0xff${widget.supportedCredentialList.elementAt(index).display.elementAt(index).textColor.toString().replaceAll('#', '')}'
+                                .isNotEmpty
```
vkubiv, line 71: this construction `supportedCredentialList.elementAt(index).display.elementAt(index)` looks realy wiered.  Each credential in this credential list contains display data for all other credentials. In another case, it will not work. 

Also, the type of supportedCredentialList is List<SupportedCredentials>. But SupportedCredentials already list of credentials. So that wrapping list is not needed. Or `SupportedCredentials` should be changed to `SupportedCredential` and contains data only for one credential.

vkubiv (changes requested):

demo/app/lib/views/supported_credentials_list.dart
talwinder50: @DRK3  Please confirm the display list is an array itself does it contain display for all other supported credentials as well ? In my opinion it only contains display for the specific supported credential 

```
 [
{
 format: jwt_vc_json, 
types: [VerifiableCredential, VerifiedEmployee], 
display: [
 { 
name: Verified Employee,
 locale: en-US, 
logo: logo.png, 
textColor: #FFFFFF, 
backgroundColor: #0000FF 
}
] 
}
]
```

vkubiv (approved):

rolsonquadras (approved):

PR #552, opened 2023-08-15
feat(sdk): Wasm integration tests.

vkubiv:


cmd/wallet-sdk-js/karma.conf.js
```diff
@@ -0,0 +1,78 @@
+// Karma configuration
+// Generated on Fri Aug 11 2023 17:00:08 GMT+0300 (Eastern European Summer Time)
```
birtony, line 2: Should be replaced with the copyright header

.github/workflows/build.yml
```diff
@@ -87,6 +87,31 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
```
birtony, line 100: Why don't we use the latest stable version (`v3`)?

cmd/wallet-sdk-js/package.json
```diff
@@ -8,7 +8,7 @@
     "build:assets": "bash scripts/build_assets.sh",
     "build": "rm -rf dist/* && npm run build:assets && npx webpack --config ./webpack.config.js --mode=production",
     "prepare": "npm run build",
-    "test": "karma start --single-run"
+    "test": "karma start --single-run -- --foo"
```
birtony, line 11: What do we need the `-- --foo` arguments for?

cmd/wallet-sdk-js/test/openid4ci.js
```diff
@@ -0,0 +1,142 @@
+import '../src/wasm_exec.js';
+import Agent from '../src/index.js';
+
+
+describe("A suite is just a function", function () {
+
+    it("and so is a spec", async function () {
```
birtony, line 7: These should be probably describe what is being tested

cmd/wallet-sdk-js/test/openid4ci.js
```diff
@@ -0,0 +1,142 @@
+import '../src/wasm_exec.js';
```
birtony, line 1: Copyright header missing

cmd/wallet-sdk-js/test/openid4ci.js
```diff
@@ -0,0 +1,142 @@
...
+            credentials: credentials,
+        });
+
+        expect(requirements.length).toBe(1);
+        expect(requirements[0].descriptors.length).toBe(1);
+        expect(requirements[0].descriptors[0].matchedVCs.length).toBe(1);
+
+        await openID4VPInteraction.presentCredential({
+            credentials: requirements[0].descriptors[0].matchedVCs
+        })
```
birtony, line 57: Might be cleaner to split the issuance and the presentation tests into 2 separate `it` clauses and save the `credentials` array globally in this file. Besides, any variables (`let`s mostly) that don't need to be changed after declaration can be changed to `const`s

scripts/wasm_test.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+#
+# Copyright Avast Software. All Rights Reserved.
```
birtony, line 3: This is an outdated copyright header

birtony (commented):
Looks good, just left a few minor comments

cmd/wallet-sdk-js/karma.conf.js
vkubiv: done

cmd/wallet-sdk-js/package.json
vkubiv: removed

cmd/wallet-sdk-js/test/openid4ci.js
vkubiv: rewrited

cmd/wallet-sdk-js/test/openid4ci.js
vkubiv: done

cmd/wallet-sdk-js/test/openid4ci.js
vkubiv: Const is fixed. 

Splitting into two sub-tests has some downsides. 
1. Changing the order of tests will break them.
2. It is not obvious to the reader of the code that one test depends on another test. 
3. We use the same single test approach for go, flutter, ios , and android test. This will be inconsistent

birtony (approved):

rolsonquadras (approved):

PR #557, opened 2023-08-17
refactor(sdk): move storageutil into wallet-sdk.

vkubiv:


go.mod
```diff
@@ -17,9 +17,9 @@ require (
 	github.com/hyperledger/aries-framework-go/component/didconfig v0.0.0-20230622211121-852ce35730b4
 	github.com/hyperledger/aries-framework-go/component/kmscrypto v0.0.0-20230622082138-3ffab1691857
 	github.com/hyperledger/aries-framework-go/component/models v0.0.0-20230718141804-b2fdf798c242
-	github.com/hyperledger/aries-framework-go/component/storageutil v0.0.0-20230427134832-0c9969493bd3
 	github.com/hyperledger/aries-framework-go/component/vdr v0.0.0-20230622171716-43af8054a539
 	github.com/hyperledger/aries-framework-go/spi v0.0.0-20230615141038-5d444d6c36de
+	github.com/hyperledger/aries-framework-go/test/component v0.0.0-20220322085443-50e8f9bd208b
```
rolsonquadras, line 22: Looks like we added a new aries component in the dependencies.

vkubiv: Yes, this is a dependency from mem store test. It will be removed after storage refactoring.

DRK3 (approved):
I thought of a way to do this in the interim that may be a bit cleaner. Consider the following:

1. (Optional) - Add a new, minimal interface to the doc loader code in Aries (where the methods still match the giant Aries storage interface methods) and use that. That should help reduce the size of the file you have to copy over. (It's better to avoid committing anything unnecessary, since it'll be in the commit history forever)
2. Instead of putting the aries storage implementation you copied over in a "legacy" folder, you could just put it directly into a folder where the final, proper storage implementation will go. Then, we we've finished the other refactoring, we just update our implementation to match. We can start figuring out our storage structure now, basically.

I'm OK with merging it regardless, though :)

rolsonquadras (approved):

PR #562, opened 2023-08-21
fix(app): Connect issuer page UI Display

talwinder50:


talwinder50:
@DRK3 @rolsonquadras  As we discussed, we were thinking it's not appropriate to make a backend call to launch the issuer display, as current method to fetch the issuer metadata is attached to interaction.  However, currently we are only initiating the interaction object after user clicks CONNECT not before that. If we need to leverage to make the backend call we would need to expose an API to fetch the issuer metadata without being tied to interaction. 

As per user experience, it doesn't make sense to depend this feature on backend call. Earlier the walletintiatedconfig was added to add theme display features as well in the config.

rolsonquadras (approved):

DRK3 (approved):

PR #563, opened 2023-08-22
chore(sdk): update to use trustbloc libs.

vkubiv:


cmd/wallet-sdk-gomobile/api/jsonwebkey.go
```diff
@@ -9,7 +9,7 @@ package api
 import (
 	"fmt"
 
-	"github.com/hyperledger/aries-framework-go/component/kmscrypto/doc/jose/jwk"
+	"github.com/trustbloc/kms-crypto-go/doc/jose/jwk"
```
troyronda, line 12: Should be trustbloc/kms-go (the repo was renamed).

vkubiv: Yes, it will be renamed. But first github.com/trustbloc/kms-go package should be changed. It will be done in tomorrow's PRs.

vkubiv: Also, we have a discussion about what to do with that kms-go repo. There few things that can affect how we actually will split aries.

DRK3 (approved):

cmd/wallet-sdk-gomobile/api/jsonwebkey.go
rolsonquadras: This will be done in follow up PR.

PR #587, opened 2023-09-05
fix(app): fixed issuance preview screen in web

birtony:

### DIFF
--- a/demo/app/lib/views/issuance_preview.dart
+++ b/demo/app/lib/views/issuance_preview.dart
@@ -40,8 +40,8 @@ class IssuancePreviewState extends State<IssuancePreview> {
   String backgroundColor = '';
   String issuerDisplayURL = '';
   String textColor = '';
-  String? logoURL;
-  String? issuerLogoURL;
+  String logoURL = '';
+  String issuerLogoURL = '';
 
   @override
   void initState() {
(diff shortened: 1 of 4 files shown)


rolsonquadras (approved):

demo/app/lib/views/issuance_preview.dart
```diff
@@ -51,9 +51,15 @@ class IssuancePreviewState extends State<IssuancePreview> {
         credentialIssuer = response.first.credentialIssuer;
         issuerDisplayName = response.first.localizedIssuerDisplays.first.name;
         issuerDisplayURL = response.first.localizedIssuerDisplays.first.url;
-        issuerLogoURL = response.first.localizedIssuerDisplays.first.logo;
+        final issuerLogo = response.first.localizedIssuerDisplays.first.logo;
+        if (issuerLogo != null) {
+          issuerLogoURL = issuerLogo;
```
DRK3, line 56: Two questions:

What happens if ultimately, there are no logos? Does the UI have some sort of placeholder image, will the UI bug out, etc?

Also, you should be able to clean up this logic here a bit.

Is `response.first.localizedIssuerDisplays.first.logo` a logo, or is it a URL to a logo? If it's a logo URL, then you should rename that `logo` field to be `logoURL` to make it clear.

Based on the null check, I'm assuming that `response.first.localizedIssuerDisplays.first.logo` could be missing, which means that `String?` would be the appropriate type for `logoURL`. There's no reason to do a null check, just to allow `logoURL` to potentially be a non-nullable-but-still potentially empty string after all.

DRK3: (This comment applies to both of the logos I see here, although my examples above are about the first issuer logo)

birtony: Good questions @DRK3. For the first one, yes, this logic is already in place. `errorWidget` renders backup logos. As for your second question, the logic that @talwinder50 put into the app before just used `logo` as a URL (a string), even though the backend returns on object. My PR fixes a very specific issue and further enhancements can be scoped and addressed later, if deemed necessary.

DRK3: I understand not wanting to rename the existing field since that's not specific to your change, but I do see you changed the type of `logoURL` and `issuerLogoURL` from a `String?` to a `String`. That looks suspicious, since the underlying type is indeed a `String?`. What's the reason for that change?

birtony: Those 2 variable cannot actually be `null` (thus, they are 'required', not optional). Otherwise, the `CachedNetworkImage` will break (as per documentation, that component takes the URL as a required parameter). For this reason, these 2 variables have a default value of an empty string over `null`. And we only assign the values fetched through the API if those are not null. I think this makes sense to do this transition here over the API layer since it is purely a UI logic.

DRK3: Makes sense @birtony , thanks for the explanation! My main goal was to check to make sure there wasn't some other sort of underlying issue that may have been buried by the `String?`->`String` conversion :)

DRK3 (approved):

PR #590, opened 2023-09-06
fix(app): fixed data handling logic in verification flow for web

birtony:


DRK3 (approved):

talwinder50 (commented):
Hi did we tested the change in the app ? So we removed the stored credential while presenting credential now ?

birtony:
> Hi did we tested the change in the app ? So we removed the stored credential while presenting credential now ?

Hey @talwinder50 ! Yes, I tested on web and on iOS as well. The parameter you're talking about was passed down but never actually used in the web logic, so I dropped it. cc @DRK3

rolsonquadras (approved):

PR #592, opened 2023-09-07
fix(app): Add exception handling for the QR code scanning process

talwinder50:
If you enter any random string in the qr code scanning text field in the simulator, that is handled now as well. 

### DIFF
--- a/demo/app/ios/Runner/OpenID4CI.swift
+++ b/demo/app/ios/Runner/OpenID4CI.swift
@@ -15,21 +15,29 @@ public class OpenID4CI {
     
     private var initiatedInteraction: Openid4ciIssuerInitiatedInteraction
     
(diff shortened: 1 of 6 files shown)


rolsonquadras (approved):

demo/app/ios/Runner/OpenID4CI.swift
```diff
@@ -55,7 +63,7 @@ public class OpenID4CI {
     
         let authorizationLink =  initiatedInteraction.createAuthorizationURL(clientID, redirectURI: redirectURI, opts: opts, error: &error)
         if let actualError = error {
-            print("error in authorizations", error!.localizedDescription)
+            print("error in creating authorization link", error!.localizedDescription)
```
DRK3, line 66:
```suggestion
            print("error while creating authorization link", error!.localizedDescription)
```

DRK3:
> Handle the crash happening while qr code

I think your commit & PR title are missing some words. I think what you mean is something like:

> Add exception handling during the QR code scanning process

?

DRK3 (approved):

PR #598, opened 2023-09-11
feat(sdk): Improved did:key create function

DRK3:
Added a new function for creating did:key DIDs. This new standalone function directly takes in a key instead of a keyWriter or keyReader, giving the caller more flexibility in terms of key management. It also minimizes input parameters to only what's required, helping reduce confusion caused by unused parameters.

### DIFF
--- a/cmd/wallet-sdk-gomobile/didkey/didkey.go
+++ b/cmd/wallet-sdk-gomobile/didkey/didkey.go
@@ -1,0 +1,38 @@
+/*
+Copyright Gen Digital Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package didkey contains a function that can be used to create did:key documents.
+package didkey
+
+import (
+	"errors"
+
+	"github.com/trustbloc/wallet-sdk/pkg/walleterror"
+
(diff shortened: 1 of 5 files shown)


cmd/wallet-sdk-gomobile/didkey/didkey.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	var vm *did.VerificationMethod
+
+	if jwk.JWK.Crv == "Ed25519" {
+		// Workaround: when the did:key VDR creates a DID for ed25519, Ed25519VerificationKey2018 is the expected
+		// verification method.
+		publicKeyBytes, err := jwk.JWK.PublicKeyBytes()
+		if err != nil {
+			return nil, wrapper.ToMobileError(err)
```
rolsonquadras, line 38: Should these be in go pkg rather than mobile binding pkg ?

DRK3: I was also wondering the same thing. I wasn't sure if we wanted this behaviour to be common to all bindings that use the Go SDK or if it should be specific to the binding implementation (gomobile, WASM). Since you were wondering the same thing, I think perhaps it makes more sense for it to be common, so I'll update the PR.

PR #614, opened 2023-09-14
feat(app): Add support for DI proofs

talwinder50:

### DIFF
--- a/demo/app/ios/Runner/OpenID4CI.swift
+++ b/demo/app/ios/Runner/OpenID4CI.swift
@@ -12,13 +12,15 @@ public class OpenID4CI {
     private var didResolver: ApiDIDResolverProtocol
     private var crypto: ApiCryptoProtocol
     private var activityLogger: ApiActivityLoggerProtocol
+    private var kms: LocalkmsKMS
     
     private var initiatedInteraction: Openid4ciIssuerInitiatedInteraction
     
-    init (requestURI: String, didResolver: ApiDIDResolverProtocol, crypto: ApiCryptoProtocol, activityLogger: ApiActivityLoggerProtocol) throws {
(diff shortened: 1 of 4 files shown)


demo/app/ios/Runner/OpenID4CI.swift
```diff
@@ -12,13 +12,15 @@ public class OpenID4CI {
     private var didResolver: ApiDIDResolverProtocol
     private var crypto: ApiCryptoProtocol
     private var activityLogger: ApiActivityLoggerProtocol
+    private var kms: LocalkmsKMS
```
birtony, line 15: `LocalkmsKMS` is a rough naming...is there anything we can do it avoid repeating ourselves?

talwinder50: Unfortunately no, this is how ios binding is generated @DRK3 have tried the options but this is how the naming is for ios binding

birtony (approved):

demo/app/ios/Runner/OpenID4CI.swift
DRK3: Yes indeed, this one was tricky. For iOS bindings, gomobile uses the Go package name as a suffix for all symbols from that package. The package name is `localkms` and the type exposed from that package is `KMS`, hence why we get `LocalkmsKMS` (gomobile capitalizes it automatically). The tricky thing is choosing package+type names that look reasonable for iOS + Android + the Go source code. I managed to improve some of the other packages with careful name choices, but at the time I didn't have a great solution for this one. Any ideas @birtony?

DRK3:
Typo in the commit title: it says `andoird` instead of `android`.

It's also missing the word "apps" at the end of the message, which is a bit confusing - support was added in the ios and android...what?

My suggestion to address that and avoid running over the character limit would be a message similar to the following:

`feat(app): Add support for DI proofs`

(the "app" part at the beginning means you don't really need to specify it again later in the mesage)

DRK3:
Also, don't forget to update the commit message too.

rolsonquadras (approved):

PR #675, opened 2023-11-21
fix(app): Push crash fix for non pin flow

talwinder50:


demo/app/lib/views/credential_preview.dart
```diff
@@ -41,6 +41,8 @@ class CredentialPreviewState extends State<CredentialPreview> {
       setState(() {
         if (response.first.issuerName.isNotEmpty){
           issuerDisplayData = response.first.issuerName;
+        } else {
+          issuerDisplayData = response.first.issuerName;
         }
```
rolsonquadras, line 46: Content inside if and else condition is same. Do we need this if-else ?

rolsonquadras (approved):

PR #691, opened 2023-12-06
feat(app): Implement android plugin changes for acknowledgement

talwinder50:
- fixes Js binding compilation issue


demo/app/android/app/src/main/kotlin/walletsdk/openid4ci/OpenID4CI.kt
```diff
@@ -123,4 +123,14 @@ class OpenID4CI constructor(
         return newInteraction.issuerMetadata()
     }
 
+    fun requireAcknowledgment(): Boolean {
+        return newInteraction.requireAcknowledgment()
+    }
+    fun acknowledgeSuccess(): Boolean {
```
talwinder50, line 129: @vkubiv  I have a question for this, 

In ios binding i do see the return type of these function is boolean but in android binding 

`- (BOOL)acknowledgeReject:(NSError* _Nullable* _Nullable)error;
/**
 * AcknowledgeSuccess acknowledges the issuer that the client accepted credentials.
 */
- (BOOL)acknowledgeSuccess:(NSError* _Nullable* _Nullable)error;`


Android binding :  Its void return type 

    public native void acknowledgeReject() throws Exception;

    public native void acknowledgeSuccess() throws Exception;


Can you please verify if its intentional or need to be fixed

PR #870, opened 2025-04-03
fix(app): CI Failure fix

talwinder50:
parallel execution
simplified YAML structure
only generates what's needed


talwinder50:
closing

PR #872, opened 2025-04-17
fix(app): CI Failure fix

talwinder50:


talwinder50:
Ios test dont have a problem, The problem for ios failure is the resource constraints. Test passes locally and some time in mac os

PR #874, opened 2025-04-25
chore(app): Migrate breaking changes in android

talwinder50:


talwinder50:
this is not fixable at this point

